PR #3, opened 2018-11-30
Add NodeInfo Model

jjbuchan:
# Related PRs
https://github.com/racker/salus-telemetry-etcd-adapater/pull/1
https://github.com/racker/salus-telemetry-ambassador/pull/3
https://github.com/racker/salus-telemetry-model/pull/3

# What
An object to store general info about an envoy.

# Why

This is what will be stored under `/nodes/expected/...` and `/nodes/active/...` for the presence monitoring stuff.

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/model/NodeInfo.java
+++ b/src/main/java/com/rackspace/salus/telemetry/model/NodeInfo.java
@@ -1,0 +1,49 @@
+/*
+ *    Copyright 2018 Rackspace US, Inc.
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ *
+ *
+ */
+
+package com.rackspace.salus.telemetry.model;
+
+import javax.validation.constraints.NotBlank;
+import javax.validation.constraints.NotNull;
+import lombok.Data;
+
+import java.util.Map;
+
+@Data
+public class NodeInfo {
+    @NotBlank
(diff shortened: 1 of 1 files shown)


src/main/java/com/rackspace/salus/telemetry/model/NodeInfo.java
```diff
@@ -0,0 +1,47 @@
...
+    Map<String, String> labels;
+
+    @NotNull
+    String envoyId;
+
+    @NotBlank
+    String tenantId;
+
+    @NotBlank
+    SocketAddress address;
```
itzg: I would have been worried that JSON conversion of `SocketAddress` may not work, but I guess it must be since it is actually being used in the other modules.

jjbuchan: Yeah, I was kinda surprised by that too.  

Here's an example of what was in etcd

```
/nodes/active/6fa62b080764684d40ba0e029b17d102b54fd313e215062b1b9f319536532eee
{"identifier":"arch","identifierValue":"X86_64","labels":{"os":"DARWIN","hostname":"MV906GG8WL","arch":"X86_64"},"envoyId":"thisisatest","tenantId":"aaaaaa","address":"[0:0:0:0:0:0:0:1]:55851"}
```

GeorgeJahad: @jjbuchan Did you try deserializing it?

This is the error I get:
```
com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `java.net.SocketAddress` (no Creators, like default construct, exist): abstract types either need to be mapped to concrete types, have custom deserializer, or contain additional type information
 at [Source: (byte[])"{"identifier":"os","identifierValue":"LINUX","labels":{"os":"LINUX","arch":"X86_64"},"envoyId":"abcde","tenantId":"123456","address":"MNL1BHG3QR:1234"}"; line: 1, column: 134] (through reference chain: com.rackspace.salus.telemetry.model.NodeInfo["address"])
```
caused by this:
```
        nodeInfo = objectMapper.readValue(kv.getValue().getBytes(), NodeInfo.class);
```

jjbuchan: Fixed in https://github.com/racker/salus-telemetry-etcd-adapter/pull/3 and https://github.com/racker/salus-telemetry-model/pull/5

PR #5, opened 2018-12-05
Fix NodeInfo serialization

jjbuchan:
Cannot use an abstract class in these models.  Casting to `InetSocketAddress` works fine.

Fixed up the annotations as well as they were wrongly swapped for two values.

Other thoughts:
Yesterday I was thinking we might want to create these objects without yet seeing a connection, so I was going to change the address field to `Nullable`, but then that'd mean the envoyId should also be able to be null.  Decided not to do that yet until we know since it'd involve extra validations in other parts of the system.

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/model/NodeInfo.java
+++ b/src/main/java/com/rackspace/salus/telemetry/model/NodeInfo.java
@@ -22,7 +22,7 @@ import javax.validation.constraints.NotBlank;
 import javax.validation.constraints.NotNull;
 import lombok.Data;
 
-import java.net.SocketAddress;
+import java.net.InetSocketAddress;
 import java.util.Map;
 
 @Data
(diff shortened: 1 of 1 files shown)


src/main/java/com/rackspace/salus/telemetry/model/NodeInfo.java
```diff
@@ -36,12 +36,12 @@
...
+    @NotBlank
     String envoyId;
 
     @NotBlank
     String tenantId;
 
-    @NotBlank
-    SocketAddress address;
+    @NotNull
+    InetSocketAddress address;
```
itzg: Now I'm wondering if we should just introduce our own simple "HostPort" data type.

jjbuchan: I'll merge for now since it's easy to change after.  This'll get things passing.

PR #15, opened 2019-01-02
Add resource crud functionality

jjbuchan:
# What

Adds some new `Resource` Entities and a CRUD Repository.

# Why

Allows us to perform crud operations on these objects.  We can now store them in SQL.

I purposely created a new Resource object to avoid interfering with the ResourceInfo that is currently stored in etcd.  Naming it `Resource` in the long run is probably better anyway.

# Testing
I've been testing manually with the ambassador, but this will be more relevant as we start to create the resource registration service.

When an envoy connects I can see this data:

```
mysql> select * from resources;
+----+-----------------+------------------+-----------+
| id | identifier_name | identifier_value | tenant_id |
+----+-----------------+------------------+-----------+
|  1 | hostname        | MV906GG8WL       | aaaaaa    |
+----+-----------------+------------------+-----------+
1 row in set (0.00 sec)

mysql> select * from labels;
+----+------------+------------+
| id | labels     | labels_key |
+----+------------+------------+
|  1 | DARWIN     | os         |
|  1 | MV906GG8WL | hostname   |
|  1 | X86_64     | arch       |
+----+------------+------------+
3 rows in set (0.00 sec)

```

# TODO

Eventually this repo should only consist of the data stored in SQL.  This gives us a good distinction for what goes in here vs. what might be stored in `etcd-adapter`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,8 +18,8 @@
   -->
 
 <project xmlns="http://maven.apache.org/POM/4.0.0"
-  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
   <groupId>com.rackspace.salus</groupId>
@@ -49,6 +49,12 @@
       <groupId>javax.validation</groupId>
       <artifactId>validation-api</artifactId>
       <version>${javax-validation.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.data</groupId>
+      <artifactId>spring-data-commons</artifactId>
+      <version>2.0.8.RELEASE</version>
+      <scope>compile</scope>
     </dependency>
 
   </dependencies>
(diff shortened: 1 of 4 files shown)


jjbuchan:
I have a `src/main/java/hibernate.cfg.xml` file locally, but it doesn't seem like it's needed.

pom.xml
```diff
@@ -18,54 +18,60 @@
   -->
 
 <project xmlns="http://maven.apache.org/POM/4.0.0"
-  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
jjbuchan: This didn't like me copy/pasting from github after the rebase conflict.  I'll deal with this file later once I verify everything else is still good.  I didn't have the `protocol` changes before pushing this.

src/main/java/com/rackspace/salus/telemetry/model/Resource.java
```diff
@@ -0,0 +1,50 @@
...
+public class Resource {
+    @Id
+    @GeneratedValue
+    Long id;
+
+    @Valid
+    ResourceIdentifier resourceIdentifier;
+
+    @ElementCollection
+    @CollectionTable(name="labels", joinColumns = @JoinColumn(name="id"))
```
itzg: I wonder if this join-like table will need to be named more specifically such as `resources_labels`. Since we'll have labels hanging off of some of our other data types, like Envoy and Monitor, we would need to name the tables distinctly since the `id` column of each refers back to their respective "owning" entity.

jjbuchan: I'm not sure, but it's a good idea to be more specific now just in case.  I'll change it.

jjbuchan:
@itzg i changed that one name.  Considered making some other tweaks but uncertain for now so will leave things until we're farther along.

An approval would be nice.  Realized I now need this for some other PRs :)

PR #17, opened 2019-01-09
Add AttachEvent which will be sent over kafka

jjbuchan:
First iteration of the attach event which will be utilized by a few services.

A thought: should we keep the sql specific data in `model` and things that will just be kafka events in a different package?

Also we might want to break away from using `InetSocketAddress`.

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/events/AttachEvent.java
+++ b/src/main/java/com/rackspace/salus/telemetry/events/AttachEvent.java
@@ -14,16 +14,16 @@
  * limitations under the License.
  */
 
-package com.rackspace.salus.telemetry.model;
+package com.rackspace.salus.telemetry.events;
 
-import lombok.Data;
-
+import com.rackspace.salus.common.messaging.KafkaMessageKey;
+import java.util.Map;
 import javax.validation.constraints.NotBlank;
 import javax.validation.constraints.NotEmpty;
 import javax.validation.constraints.NotNull;
-import java.net.InetSocketAddress;
-import java.util.Map;
+import lombok.Data;
 
+@KafkaMessageKey(properties = {"tenantId", "identifierName", "identifierValue"})
 @Data
 public class AttachEvent {
     @NotBlank
@@ -44,6 +44,11 @@ public class AttachEvent {
     @NotBlank
     String tenantId;
 
+    /**
+     * The remote hostname or IP address of the Envoy's TCP connection to the Ambassador.
+     * Note that this may be a misleading address if any NAT or proxy is used along the path of
+     * the connection.
+     */
     @NotNull
-    InetSocketAddress address;
+    String envoyAddress;
 }
--- a/src/main/java/com/rackspace/salus/telemetry/model/AttachEvent.java
+++ b/src/main/java/com/rackspace/salus/telemetry/model/AttachEvent.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2019 Rackspace US, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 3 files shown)


src/main/java/com/rackspace/salus/telemetry/model/AttachEvent.java
```diff
@@ -0,0 +1,49 @@
...
+import lombok.Data;
+
+import javax.validation.constraints.NotBlank;
+import javax.validation.constraints.NotEmpty;
+import javax.validation.constraints.NotNull;
+import java.net.InetSocketAddress;
+import java.util.Map;
+
+@Data
+public class AttachEvent {
```
itzg: Not directly related to this PR, but I'm now wondering, what does the resource manager expect during an Envoy detachment. Should Ambassador send anything? If it should send something, I would propose it send a "tombstone" message with the same message key as this AttachEvent knowing that a `KafkaListener` can be confirmed to receive those null/tombstones:

https://docs.spring.io/spring-kafka/docs/2.2.2.RELEASE/reference/html/_reference.html#tombstones

itzg: Along the lines of the tombstone question, I would like to propose that in the class javadoc of these events that we document what the kafka message key structure should be. In this case, I'm thinking `{tenantId}:{identiferName}:{identifierValue}`.

In fact, we could craft a custom annotation and a helper service to do that declaration and field extraction+key building, respectively. I'll push a PR to show what I mean...but not sure which repo to put it in yet :)

jjbuchan:
> Thought I'd approve in case you wanted to merge and then I take over remaining tweaks like the address

I was thinking this would stay here as the working version and we can each make tweaks as desired to come to the final solution.  Can just pull in the branch locally until we've settled on the final object.

itzg:
I confirmed that the model->common dependency I added doesn't introduce any cycles or dependency awkwardness:

salus telemetry module dependencies

src/main/java/com/rackspace/salus/telemetry/events/AttachEvent.java
```diff
@@ -0,0 +1,54 @@
...
+    @NotBlank
+    String tenantId;
+
+    /**
+     * The remote hostname or IP address of the Envoy's TCP connection to the Ambassador.
+     * Note that this may be a misleading address if any NAT or proxy is used along the path of
+     * the connection.
+     */
+    @NotNull
+    String envoyAddress;
```
itzg: @jjbuchan with this change I'm proposing that we just drop the tracking of the remote port of the Envoy's connection. It'll just be an ephemeral port, so it doesn't add much value for diagnosing things.

src/main/java/com/rackspace/salus/telemetry/model/AttachEvent.java
jjbuchan: Unless we want to remove the envoy labels everytime an envoy disconnects I don't think we need to send anything.  That is a thought though... do we want envoy labels lingering around on a resource if it hasn't been active in a long time?  Does it hurt anything to leave them there?

src/main/java/com/rackspace/salus/telemetry/events/AttachEvent.java
jjbuchan: works for me

src/main/java/com/rackspace/salus/telemetry/model/AttachEvent.java
itzg: Agreed, it doesn't feel like it hurts to leave the "last known labels" on the resource indefinitely. If anything it could be interesting to look at what those were.

PR #19, opened 2019-01-15
Refactor KafkaMessageType

itzg:
# Resolves

Part of SALUS-136

# What

Moves the KafkaMessageType from ambassador's "types" package, but aligns the changes made in the resource-manager. I also renamed the "events" package to "messaging" to align with the convention in the common module, but also better name the package to allow for all things related to the Kafka messaging.

# TODO

There will be an update to the ambassador PR https://github.com/racker/salus-telemetry-ambassador/pull/14


GeorgeJahad:
looks good to me

PR #22, opened 2019-01-16
Add resource events and set resourceId

jjbuchan:
# What

It adds the event that will be sent via Kafka whenever a resource change occurs.  Also modifies the existing resource object a little.

We made a decision to use `resourceId` instead of the `identifierName/Value` combo.  This includes those changes.

## How to test

This will be part of the Resource Management service PR.

# Why

The Monitor Management and Presence Monitor services will need to consume these.

# TODO

Finalize interaction between resource management and monitor management.

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/messaging/ResourceEvent.java
+++ b/src/main/java/com/rackspace/salus/telemetry/messaging/ResourceEvent.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2019 Rackspace US, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.rackspace.salus.telemetry.messaging;
+
+import java.util.Map;
+import javax.validation.constraints.NotBlank;
+import javax.validation.constraints.NotNull;
+
+import com.rackspace.salus.common.messaging.KafkaMessageKey;
+import com.rackspace.salus.telemetry.model.Resource;
+import lombok.Data;
+
+@KafkaMessageKey(properties = {"tenantId", "operation"})
+@Data
+public class ResourceEvent {
+    @NotBlank
+    Resource resource;
+
+    @NotNull
+    Map<String,String> oldLabels;
+
+    @NotBlank
+    String operation;
+
+    @NotNull
+    boolean presenceMonitorChange;
+
+    @NotBlank
+    String tenantId;
+}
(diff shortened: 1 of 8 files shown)


src/main/java/com/rackspace/salus/telemetry/messaging/ResourceEvent.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.Map;
+import javax.validation.constraints.NotBlank;
+import javax.validation.constraints.NotNull;
+
+import com.rackspace.salus.common.messaging.KafkaMessageKey;
+import com.rackspace.salus.telemetry.model.Resource;
+import lombok.Data;
+
+@KafkaMessageKey(properties = {"tenantId", "operation"})
```
jjbuchan: Wasn't entirely sure what the key of this should be since all the resource properties are within the resource object.  Went with this to start with until we figure out what we need/want.

itzg: Since the message key drives partitioning and the app instance of the consumer group that consumes the event, then it shouldn't include `operation` since we'd want events of the same tenant to route to the same consumer.

src/main/java/com/rackspace/salus/telemetry/messaging/ResourceEvent.java
```diff
@@ -0,0 +1,44 @@
...
+@Data
+public class ResourceEvent {
+    @NotBlank
+    Resource resource;
+
+    @NotNull
+    Map<String,String> oldLabels;
+
+    @NotBlank
+    String operation;
```
GeorgeJahad: shouldn't this be an enum?

src/main/java/com/rackspace/salus/telemetry/messaging/ResourceEvent.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @NotNull
+    boolean presenceMonitorChange;
+
+    @NotBlank
+    String tenantId;
```
GeorgeJahad: since this is already in the resource, is it not a duplicate here?

GeorgeJahad:
originally, we were thinking  the getAllResources() rest api would take a "hasEnvoy" parameter so as to only return those resources that do have an envoy.  is that still the plan?

src/main/java/com/rackspace/salus/telemetry/messaging/ResourceEvent.java
jjbuchan: Yeah, was just using string to get me started.  Was going to do it the same as `KafkaMessageType`

src/main/java/com/rackspace/salus/telemetry/messaging/ResourceEvent.java
jjbuchan: Can't remember all the reasons why I did it like that, but thoughts of the kafka key was definitely one of them.

jjbuchan:
Yep, I have this set in the API but haven't focused on that section yet to finalize the logic and make it like what we discussed.

```
    @GetMapping("/envoys")
    public Stream getAllWithEnvoys() {
        return resourceManagement.getResources(true);
    }
```

src/main/java/com/rackspace/salus/telemetry/messaging/ResourceEvent.java
jjbuchan: Fixed

PR #23, opened 2019-01-21
Consolidating down the resource identifier concept into a single field

itzg:
# Resolves

SALUS-148

# What

Consolidating the identifier name/value fields down into a single `resourceId` field.

## How to test

Unit tests are updated in the respective users of this library.

# TODO

There will be related PRs in etcd-adapter and ambassador


itzg:
It looks like all of my changes are a duplicate of https://github.com/racker/salus-telemetry-model/pull/22

PR #26, opened 2019-02-16
added potential MonitorEvent Data Object

harageth:
# Resolves

    SALUS-129

# What

    add a sharable data object for MonitorEvents that can also be used in the Ambassador

# How

    I tried to grab what looks like the data that the Ambassador is already grabbing from etcd. 

## How to test

    <instructions for verifying the changes specific to this PR>

# Why

    This data seems similar to the ResourceEvent. Depending on the circumstance we might even be able to use the ResourceEvent. Or just a Resource Object. But the Ambassador code seems to be written to take a list of multiple different OperationTypes.

# TODO

    This is a WIP so I need consensus from someone else at least.

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/messaging/MonitorEvent.java
+++ b/src/main/java/com/rackspace/salus/telemetry/messaging/MonitorEvent.java
@@ -1,6 +1,8 @@
 package com.rackspace.salus.telemetry.messaging;
 
 import java.util.Map;
+
+import com.rackspace.salus.telemetry.model.AgentConfig;
 import lombok.Data;
 
 @Data
@@ -9,14 +11,13 @@ public class MonitorEvent {
     String ambassadorId;
     String envoyId;
 
-    //checkData
-    Map<String, String> labels;
-
     //tenantOfEnvoy
     String tenantId;
 
     OperationType operationType;
 
     //Optional: for remote checks
-    String customerTenantTag;
+    String targetTenant;
+
+    AgentConfig config;
 }
--- a/src/main/java/com/rackspace/salus/telemetry/messaging/MonitorEvent.java
+++ b/src/main/java/com/rackspace/salus/telemetry/messaging/MonitorEvent.java
@@ -1,0 +1,16 @@
+package com.rackspace.salus.telemetry.messaging;
+
+import java.util.Map;
+import lombok.Data;
+
+@Data
+public class MonitorEvent {
+
+    public boolean presenceMonitoringEnabled;
+
+    String tenantId;
+
+    String resourceId;
+
+    public Map<OperationType, Map<String, String>> operationOrganizedLabels;
+}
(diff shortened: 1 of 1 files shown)


src/main/java/com/rackspace/salus/telemetry/messaging/MonitorEvent.java
```diff
@@ -0,0 +1,16 @@
+package com.rackspace.salus.telemetry.messaging;
+
+import java.util.Map;
+import lombok.Data;
+
+@Data
+public class MonitorEvent {
+
+    public boolean presenceMonitoringEnabled;
```
itzg: What is this for?

src/main/java/com/rackspace/salus/telemetry/messaging/MonitorEvent.java
```diff
@@ -0,0 +1,16 @@
...
+
+    String tenantId;
+
+    String resourceId;
```
itzg: Monitoring isn't applied to any one resource in Salus since it's all label driven. So, I can't think of why this would be needed.

src/main/java/com/rackspace/salus/telemetry/messaging/MonitorEvent.java
```diff
@@ -0,0 +1,16 @@
...
+
+    public Map<OperationType, Map<String, String>> operationOrganizedLabels;
```
itzg: The essence of a monitor is the actual agent (such as telegraf) configuration and the indicators of what instances should apply that config. This field should probably be an https://github.com/racker/salus-telemetry-model/blob/master/src/main/java/com/rackspace/salus/telemetry/model/AgentConfig.java#L27 with a companion `OperationType`

src/main/java/com/rackspace/salus/telemetry/messaging/MonitorEvent.java
```diff
@@ -0,0 +1,16 @@
+package com.rackspace.salus.telemetry.messaging;
+
+import java.util.Map;
+import lombok.Data;
+
+@Data
+public class MonitorEvent {
```
itzg: We'll need to check with @jjbuchan , but I was expecting some kind of Envoy identification in here to tell the consuming Ambassadors if this monitor applies to one of their managed Envoys.

itzg:
I just double checked @jjbuchan 's [diagram](https://raw.githubusercontent.com/racker/salus-telemetry-bundle/master/docs/check-creation-api.drawio.png) and step 6 outlines the same kinds of fields I suggested:

image

src/main/java/com/rackspace/salus/telemetry/messaging/MonitorEvent.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.util.Map;
+
+import com.rackspace.salus.telemetry.model.AgentConfig;
+import lombok.Data;
+
+@Data
+public class MonitorEvent {
+
+    String ambassadorId;
```
jjbuchan: I feel like these should have some annotations for things like `@NotBlank`/@`NotNull` etc.

PR #27, opened 2019-02-18
add monitor model classes

GeorgeJahad:
# Resolves

https://jira.rax.io/browse/SALUS-130
# What

model classes needed by the monitor manager

# TODO
Decide if "REMOTE" is a valid agent type or if we need some other structure.

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/errors/MonitorAlreadyExists.java
+++ b/src/main/java/com/rackspace/salus/telemetry/errors/MonitorAlreadyExists.java
@@ -1,0 +1,21 @@
+/*
+ * Copyright 2019 Rackspace US, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.rackspace.salus.telemetry.errors;
+
+public class MonitorAlreadyExists extends RuntimeException {
+    public MonitorAlreadyExists(String message) { super(message);};
+}
--- a/src/main/java/com/rackspace/salus/telemetry/model/AgentType.java
+++ b/src/main/java/com/rackspace/salus/telemetry/model/AgentType.java
@@ -20,6 +20,5 @@ package com.rackspace.salus.telemetry.model;
 
 public enum AgentType {
     TELEGRAF,
-    FILEBEAT,
-    REMOTE
+    FILEBEAT
 }
--- a/src/main/java/com/rackspace/salus/telemetry/model/AgentType.java
+++ b/src/main/java/com/rackspace/salus/telemetry/model/AgentType.java
@@ -1,5 +1,5 @@
 /*
- *    Copyright 2018 Rackspace US, Inc.
+ *    Copyright 2019 Rackspace US, Inc.
  *
  *    Licensed under the Apache License, Version 2.0 (the "License");
  *    you may not use this file except in compliance with the License.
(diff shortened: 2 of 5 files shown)


src/main/java/com/rackspace/salus/telemetry/errors/MonitorAlreadyExists.java
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.rackspace.salus.telemetry.errors;
+
+public class MonitorAlreadyExists extends RuntimeException {
```
itzg: Perhaps name this just `AlreadyExists` so we can use for other cases. While you're at it, it'll feel redundant but it's idiomatic to name exception classes ending in `Exception`, so this would become `AlreadyExistsException`.

src/main/java/com/rackspace/salus/telemetry/model/AgentType.java
```diff
@@ -20,5 +20,6 @@
 
 public enum AgentType {
     TELEGRAF,
-    FILEBEAT
+    FILEBEAT,
+    REMOTE
```
itzg: If I had actually documented the enum it would have been evident that this enum indicates a particular agent package to be managed by the Envoys. So the "remote" designation will either be a completely new field on the monitor config or implied by the selector scope being `ONE_OF`...or just by whichever agent plugin is configured.

src/main/java/com/rackspace/salus/telemetry/model/Monitor.java
```diff
@@ -0,0 +1,56 @@
...
+import java.io.Serializable;
+import java.util.Map;
+
+import static com.rackspace.salus.telemetry.model.AgentType.REMOTE;
+
+@Entity
+@Table(name = "monitors",
+        uniqueConstraints={@UniqueConstraint(columnNames={"tenant_id", "monitor_id"})})
+@Data
+public class Monitor implements Serializable {
```
itzg: May want to double check things line up with the `MonitorEvent` from https://github.com/racker/salus-telemetry-model/pull/26 . The `targetTenant` isn't immediately going to be used but is one I see.

src/main/java/com/rackspace/salus/telemetry/model/Monitor.java
```diff
@@ -0,0 +1,56 @@
...
+    Map<String,String> labels;
+
+    @NotBlank
+    @Column(name="tenant_id")
+    String tenantId;
+
+    @NotBlank
+    String content;
+
+    AgentType agentType = REMOTE;
```
itzg: There's not really any one default value other than `TELEGRAF`, but only because that's the main one being used.

GeorgeJahad:
requested changes are here:
https://github.com/racker/salus-telemetry-model/compare/85d8173804ba5e074dc92bcc1d739c2836d3cb13...george/addMonitor

src/main/java/com/rackspace/salus/telemetry/model/Monitor.java
```diff
@@ -0,0 +1,62 @@
...
+        uniqueConstraints={@UniqueConstraint(columnNames={"tenant_id", "monitor_id"})})
+@Data
+public class Monitor implements Serializable {
+    @Id
+    @GeneratedValue
+    Long id;
+
+    @NotNull
+    @Column(name="monitor_id")
+    String monitorId;
```
jjbuchan: What is an example of this field?

I'm not sure customers should be expected to specify a monitorId the same way they do for resourceId?

GeorgeJahad: so how do they retrieve them then?  with some kind of precalculated monitor id?

jjbuchan: By the generated id above?  We could change that to a uuid instead of a long to make it a little nicer.

GeorgeJahad: so you would use that for the primary index then?

jjbuchan: I might be wrong, but I think that's what the `@Id` annotation is already doing.  The `UniqueConstraint` option is separate from primary key.

jjbuchan: Yep, https://docs.oracle.com/javaee/6/api/javax/persistence/Id.html

GeorgeJahad:
I closed this one so we could finish the merges required for this sprint.  I've opened up a separate jira ticket for the issue mentioned by @jjbuchan:
https://jira.rax.io/browse/SALUS-202

PR #30, opened 2019-02-26
fix monitor model to use generated UUID for monitor retrieval

GeorgeJahad:
https://jira.rax.io/browse/SALUS-202

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -62,31 +62,9 @@
       <artifactId>validation-api</artifactId>
     </dependency>
     <dependency>
-      <!-- for compiling with Java 11 -->
-      <groupId>javax.xml.bind</groupId>
-      <artifactId>jaxb-api</artifactId>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-data-jpa</artifactId>
     </dependency>
-    <dependency>
-      <groupId>org.springframework.data</groupId>
-      <artifactId>spring-data-commons</artifactId>
-      <version>2.0.8.RELEASE</version>
-    </dependency>
-    <dependency>
-      <groupId>javax.persistence</groupId>
-      <artifactId>javax.persistence-api</artifactId>
-      <version>2.2</version>
-    </dependency>
-    <dependency>
-      <groupId>org.hibernate</groupId>
-      <artifactId>hibernate-jpamodelgen</artifactId>
-      <version>5.3.7.Final</version>
-    </dependency>
-    <dependency>
-      <groupId>org.hibernate</groupId>
-      <artifactId>hibernate-core</artifactId>
-      <version>5.2.17.Final</version>
-    </dependency>
-
   </dependencies>
 
   <dependencyManagement>
--- a/pom.xml
+++ b/pom.xml
@@ -81,6 +81,16 @@
       <artifactId>hibernate-jpamodelgen</artifactId>
       <version>5.3.7.Final</version>
     </dependency>
+      <dependency>
+          <groupId>org.hibernate</groupId>
+          <artifactId>hibernate-core</artifactId>
+          <version>5.2.17.Final</version>
+      </dependency>
+    <dependency>
+      <groupId>org.hibernate</groupId>
+      <artifactId>hibernate-core</artifactId>
+      <version>5.2.17.Final</version>
+    </dependency>
 
   </dependencies>
 
--- a/pom.xml
+++ b/pom.xml
@@ -81,11 +81,6 @@
       <artifactId>hibernate-jpamodelgen</artifactId>
       <version>5.3.7.Final</version>
     </dependency>
-      <dependency>
-          <groupId>org.hibernate</groupId>
-          <artifactId>hibernate-core</artifactId>
-          <version>5.2.17.Final</version>
-      </dependency>
     <dependency>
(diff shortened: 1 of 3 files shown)


src/main/java/com/rackspace/salus/telemetry/model/Monitor.java
```diff
@@ -25,19 +25,19 @@
...
 @Entity
-@Table(name = "monitors",
-        uniqueConstraints={@UniqueConstraint(columnNames={"tenant_id", "monitor_id"})})
+@Table(name = "monitors")
 @Data
 public class Monitor implements Serializable {
     @Id
     @GeneratedValue
-    Long id;
+    @Column(unique=true, columnDefinition = "BINARY(16)")
```
itzg: Did JPA/Hibernate/MySQL require this to be binary? This seems like it'll be awkward to manually inspect data via SQL queries.

The `unique=true` might be redundant since the field is already marked with an `@Id`, which makes it a primary key and therefore has a uniqueness constraint.

itzg: I'm almost worried still that we're working against the natural behavior of MySQL (and SQL in general) by not using an auto-increment column for primary key

https://dev.mysql.com/doc/refman/8.0/en/example-auto-increment.html

GeorgeJahad: I was unable to retrieve the record by uuid without that annotation.  There may be other alternatives, like converting it to a string, but that seemed less efficient.  I'm fine with autoincrement.

Interestingly, the in-memory db used by the DataJpaTest annotation doesn't suffer this problem, so my tests run fine without the binary(16) annotation.  Only with mysql is it a problem.

pom.xml
```diff
@@ -81,6 +81,11 @@
       <artifactId>hibernate-jpamodelgen</artifactId>
       <version>5.3.7.Final</version>
     </dependency>
+    <dependency>
+      <groupId>org.hibernate</groupId>
+      <artifactId>hibernate-core</artifactId>
```
itzg: Sorry I didn't notice this before, but you seem to be missing `spring-boot-starter-data-jpa` which is what you should be using to bring in all the compatible hibernate pieces. For example, I see this in my event management app's pom:

image

With that you can also remove the explicit dependency on at least javax.persistence-api and spring-data-commons also. You can cross-check any others here https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-starter-data-jpa/2.1.2.RELEASE

src/main/java/com/rackspace/salus/telemetry/model/Monitor.java
```diff
@@ -25,19 +25,19 @@
...
 @Entity
-@Table(name = "monitors",
-        uniqueConstraints={@UniqueConstraint(columnNames={"tenant_id", "monitor_id"})})
+@Table(name = "monitors")
 @Data
 public class Monitor implements Serializable {
     @Id
     @GeneratedValue
-    Long id;
+    @org.hibernate.annotations.Type(type="uuid-char")
```
itzg: MUCH better than the binary override.

pom.xml
itzg: ...FYI you'll want to use the "starter" dependencies as much as possible since the Spring Boot people have curated specific dependency versions via those. Here's more info about the starters:

https://docs.spring.io/spring-boot/docs/current/reference/html/using-boot-build-systems.html#using-boot-starter

src/main/java/com/rackspace/salus/telemetry/model/Monitor.java
jjbuchan: fyi, if we did ever want to store it as binary, they also allow for `uuid-binary`.  There are various hibernate mappings to go from the java data type to the type in the db.  I think for UUID the only options were char and binary though.

pom.xml
GeorgeJahad: I certainly don't mind adding it, but just to be clear this is the pom for the model repo, so adding it here will pull it in for all the other apps that use the models.   Should we remove it from all the other poms?  I guess until this annotation, there have been no jpa/hibernate specific annotations in the models.

itzg: Yeah, the model hasn't contained JPA references until the management apps. The dependencies are already "polluted" by JPA/Hibernate, so using the starter instead is no worse.

Ideally the model module would not make any JPA annotation references, but I haven't yet found a solution I liked for having a shared "generic" class, but then subclass with the JPA specifics.

itzg: ...the solution I'm trying out so far is declaring a DTO class like

https://github.com/racker/salus-event-engine-management/blob/43d2bf6067d0e49d87f9da41d74de2954851676c/src/main/java/com/rackspace/salus/event/manage/model/EventEngineTaskDTO.java#L23

that mirrors the entity class

https://github.com/racker/salus-event-engine-management/blob/43d2bf6067d0e49d87f9da41d74de2954851676c/src/main/java/com/rackspace/salus/event/manage/entities/EventEngineTask.java#L40

and using object mapper to convert between them:

https://github.com/racker/salus-event-engine-management/blob/43d2bf6067d0e49d87f9da41d74de2954851676c/src/main/java/com/rackspace/salus/event/manage/web/TasksApi.java#L62

PR #31, opened 2019-03-03
Use labels in AgentRelease instead of os and arch fields

itzg:
# Resolves

While working on https://jira.rax.io/browse/SALUS-205

# What

Makes the AgentRelease model more consistent with the Monitor model where os and arch are already labels there and should be treated the same here in AgentRelease.

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/model/Architecture.java
+++ b/src/main/java/com/rackspace/salus/telemetry/model/Architecture.java
@@ -1,24 +1,24 @@
 /*
- *    Copyright 2018 Rackspace US, Inc.
+ * Copyright 2019 Rackspace US, Inc.
  *
- *    Licensed under the Apache License, Version 2.0 (the "License");
- *    you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


src/main/java/com/rackspace/salus/telemetry/model/Architecture.java
```diff
@@ -1,24 +1,24 @@
...
  */
 
 package com.rackspace.salus.telemetry.model;
 
 public enum Architecture {
     X86,
-    X86_64
+    X86_64;
+
+    public static final String NAME = "arch";
```
GeorgeJahad: Out of curiousity, are these static "NAME"s used anywhere?

itzg: Yes, they’re used in the api code changes that I haven’t yet pushed out.

itzg:
@GeorgeJahad , I fixed merge conflicts. Can I a re-review please? (Had enabled the clear-review-on-push setting in our repos.)

PR #33, opened 2019-03-04
Add a data type which supports ObjectMapper deserializing of Page

itzg:
# Resolves

Needed for https://jira.rax.io/browse/SALUS-205

# What

The Spring Data `Page` (and its concrete class `PageImpl) that we use for returning our "get range" API methods is not deserializable by Jackson ObjectMapper due to a lack of default constructor:

```
org.springframework.http.converter.HttpMessageConversionException: Type definition error: [simple type, class org.springframework.data.domain.Page]; nested exception is com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `org.springframework.data.domain.Page` (no Creators, like default construct, exist): abstract types either need to be mapped to concrete types, have custom deserializer, or contain additional type information
```

[This SO answer](https://stackoverflow.com/a/52160387/121324) didn't work when eliminating the unused constructor parameters. It probably does work with all the parameters, but since they're discarded it seems like a brute force/ugly way to solve the problem.

# How

...instead, this PR introduces our own, simple data class that encapsulates just the primary fields we need to convey from the `Page` and it is ObjectMapper friendly.

## How to test

A unit test was added.

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/model/PagedContent.java
+++ b/src/main/java/com/rackspace/salus/telemetry/model/PagedContent.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2019 Rackspace US, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


itzg:
@GeorgeJahad I'm using it like this:

```
    final ResponseEntity<PagedContent<Monitor>> response = monitorManagementRest
        .exchange(
            "/api/tenant/{tenantId}/monitors?size={size}&page={page}",
            HttpMethod.GET,
            null,
            new ParameterizedTypeReference<PagedContent<Monitor>>() {
            },
            tenantId,
            size,
            page
        );
```

GeorgeJahad:
That helps, thanks!

PR #35, opened 2019-03-05
Added a PagedContent singleton builder

itzg:
# Resolves

Part of https://jira.rax.io/browse/SALUS-204

# What

Adds a helper method for building a `PagedContent` from a single value.

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/model/PagedContent.java
+++ b/src/main/java/com/rackspace/salus/telemetry/model/PagedContent.java
@@ -16,6 +16,7 @@
 
 package com.rackspace.salus.telemetry.model;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.function.Function;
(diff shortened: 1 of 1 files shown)


jjbuchan:
I don't really get this.  Why would we be sending paged attributes for a single value?  Wouldn't it just return data for the single value rather than a list?

itzg:
> I don't really get this. Why would we be sending paged attributes for a single value? Wouldn't it just return data for the single value rather than a list?

This is a GraphQL quirk. In order to overload the `monitors` and `resources` queries all overloaded methods must return the same data type...so a list or paged content thereof is the common return type.

jjbuchan:
Oh right.  The internal service apis dont use this at all, just graphql?

itzg:
@jjbuchan in case it helps, here's how the overloaded `monitors` looks in the schema docs:

image

PR #38, opened 2019-03-22
Declare label namespaces and utility methods for processing those

itzg:
# Resolves

Part of https://jira.rax.io/browse/SALUS-265 and https://jira.rax.io/browse/SALUS-262

# What

Declares constants for our standardized namespaces and a pair of utility methods for validating and applying those.

## How to test

New unit tests have been added

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/model/LabelNamespaces.java
+++ b/src/main/java/com/rackspace/salus/telemetry/model/LabelNamespaces.java
@@ -23,9 +23,9 @@ import org.springframework.util.Assert;
 public class LabelNamespaces {
 
   /**
-   * Scopes the labels that were discovered by an Envoy and reported during attachment
+   * Scopes the labels that were provided by an Envoy during attachment
    */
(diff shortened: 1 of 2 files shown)


src/main/java/com/rackspace/salus/telemetry/model/LabelNamespaces.java
```diff
@@ -0,0 +1,66 @@
...
+  /**
+   * Scopes the labels that were discovered by an Envoy and reported during attachment
+   */
+  public static final String DISCOVERED = "discovered";
+
+  /**
+   * Scopes the tags of the metrics ingested into the Event Engine that were fields originating
+   * from top-level system identifiers, such as account and resource ID
+   */
+  public static final String EVENT_ENGINE_TAGS = "system";
```
itzg: Regarding this comment

https://github.com/racker/salus-event-engine-ingest/pull/2#discussion_r268730885

maybe the namespace here should be `identifiers`, but now we're gun shy about that term :)

PR #39, opened 2019-04-06
update model for remote poller

GeorgeJahad:
# Resolves

part of https://jira.rax.io/browse/SALUS-290

# What
adds the bound monitor table, and the zone fields to the db.  

Part of the ticket was to also modify the event definitions to remove the sql data.  I'll do that in another PR, because that breaks a lot of things, and I want people to be able to use this without fixing everything first.

## How to test
mvn tests

# TODO
I'm adding a pr for resource and monitor management to modify the apis for the zone changes in the model


src/main/java/com/rackspace/salus/telemetry/model/BoundMonitor.java
```diff
@@ -0,0 +1,70 @@
...
+    @Column(name="tenant_id", length=100)
+    String tenantId;
+
+    @javax.persistence.Id
+    @NotNull
+    @Column(name="resource_id", length=100)
+    String resourceId;
+
+    @Column(name="resolved_template")
+    String resolvedTemplate;
```
itzg: Seeing the use of `length` attribute reminds me we should be declaring `@Lob` on this field and `content` of `Monitor`. Not that the agent config JSON will be huge, but we should allow for large-ish amounts of content.

src/main/java/com/rackspace/salus/telemetry/model/BoundMonitor.java
```diff
@@ -0,0 +1,70 @@
...
+ }
+    @javax.persistence.Id
+    @NotNull
+    @org.hibernate.annotations.Type(type="uuid-char")
+    @Column(name="monitor_id", length= 100)
+    UUID monitorId;
+
+    @javax.persistence.Id
+    @NotBlank
+    @Column(name="tenant_id", length=100)
```
itzg: Good idea with the `length`. We should eventually retrofit our other entity classes with this.

src/main/java/com/rackspace/salus/telemetry/model/BoundMonitor.java
```diff
@@ -0,0 +1,70 @@
...
+    String tenantId;
+
+    @javax.persistence.Id
+    @NotNull
+    @Column(name="resource_id", length=100)
+    String resourceId;
+
+    @Column(name="resolved_template")
```
itzg: I think it will derive the same, snake-case column name from the camel case field name...but doesn't hurt to be explicit.

src/main/java/com/rackspace/salus/telemetry/model/BoundMonitor.java
```diff
@@ -0,0 +1,70 @@
...
+    String resolvedTemplate;
+
+    @javax.persistence.Id
+    @Column(name="zone", length=100)
+    String zone;
+
+    @Column(name="envoy_id")
```
itzg: Just curious why not a `length` constraint on this one? Envoy IDs are UUIDs if that helps.

src/main/java/com/rackspace/salus/telemetry/model/Monitor.java
```diff
@@ -61,4 +61,6 @@
     @Column(name="selector_scope")
     @Enumerated(EnumType.STRING)
     ConfigSelectorScope selectorScope;
+
+    String[] zones;
```
itzg: JPA probably supports `List<String>` too, which are generally easier to work with than arrays.

src/main/java/com/rackspace/salus/telemetry/model/Resource.java
```diff
@@ -52,4 +52,6 @@
 
     @NotNull
     Boolean presenceMonitoringEnabled;
+
+    String region;
```
itzg: I was expecting this field to be called `zone`, but then the indirection of the term "region" does make sense we'd be maintaining some kind of mapping of region to default zones.

itzg:
As part of my local changes, I'm actually moving `BoundMonitor` into the `monitor-management` module since I'm now thinking it should be "private" to that module. As such don't worry about making the changes I mentioned since I have made them and validated as I moved it.

src/main/java/com/rackspace/salus/telemetry/model/BoundMonitor.java
GeorgeJahad: Actually I had to do this because mysql has a keylength limit that I was exceeding with the 4 fields at 255 each.

itzg: Ah, good to know about the key length limit.

src/main/java/com/rackspace/salus/telemetry/model/Resource.java
GeorgeJahad: Someone, (James?), said they thought it should be called region.

GeorgeJahad:
Will you be moving the boundMonitorRepository class as well?

itzg:
> Will you be moving the boundMonitorRepository class as well?

Yes, in fact it dawned on me that we shouldn't really have any repositories in the model module since repositories are a type of service that should be accessed by only one particular microservice.

src/main/java/com/rackspace/salus/telemetry/model/Resource.java
itzg: Region and zone are somewhat interchangeable terms, so I wonder if it got used verbally without purposeful distinction. I don't see any comments on the doc PR that call out region separately from zone.

GeorgeJahad: it is mentioned here, but I don't feel strongly about it: 
 https://one.rackspace.com/pages/viewpage.action?spaceKey=monitoring&title=Explicit+Poller+Picking+Design#ExplicitPollerPickingDesign-Othertasks

Do you want me to change it?

itzg: Yeah I noticed it was there which is why I searched the PR :) I'm on the fence too, but thinking we should just stick to "zone" until we find a need to differentiate that from a "region".

GeorgeJahad: it's in this pr: https://github.com/racker/ele-kb/pull/268#discussion_r270282485

itzg: Oh, I'm sorry, I was looking at what was the older of the two PRs. Cool, "region" it is then.

GeorgeJahad:
latest commit removes boundMonitor/repo, and changes zones to list: https://github.com/racker/salus-telemetry-model/pull/39/commits/a4dc6b1b30385e15b89a3c4d0dc25b243405c39a

PR #40, opened 2019-04-09
model event changes

GeorgeJahad:
# Resolves

part of https://jira.rax.io/browse/SALUS-290

# What

removes sql data from the model events, forcing consumers to read the data from the database after receiving.

# TODO

This PR breaks resourceMan, ambassador, monitorMan, and presence monitor, (the producers/consumers of these events), and should not be merged until those projects are updated as well.

I'll be submitting the fix for resourceMan shortly.

The fixes for the ambassador and monitorMan are jira tickets already defined in the remote poller epic.

If I have time before my leave, I'll fix pmon.  Otherwise I'll create a jira ticket describing what needs to be done.


itzg:
I reverted the `MonitorEvent` changes from this PR since that class is eventually going away and being replaced by the new `MonitorBoundEvent`.

PR #41, opened 2019-04-12
Introduce MonitorBoundEvent and move BoundMonitor into management app

itzg:
# Resolves

Part of 
- https://jira.rax.io/browse/SALUS-292
- https://jira.rax.io/browse/SALUS-305

# What

- Introduce MonitorBoundEvent, which will eventually replace all use of MonitorEvent since we no longer want events to convey content details
- Introduce the `REMOTE` scope on `ConfigSelectorScope` which eventually will get renamed and `ALL_OF` renamed to `LOCAL` or `AGENT`.
- Moved all JPA repositories to their respective apps since they are not "model" classes but rather services

## How to test

Unit test added


src/main/java/com/rackspace/salus/telemetry/messaging/MonitorBoundEvent.java
```diff
@@ -14,13 +14,15 @@
...
 
-import java.util.UUID;
-
-
-public interface MonitorRepository extends PagingAndSortingRepository<Monitor, UUID> {
+/**
+ * Indicates that the given envoy has one or more monitors that have changed binding
+ */
+@Data
+public class MonitorBoundEvent {
```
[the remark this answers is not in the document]
itzg: Yep, it sure is

src/main/java/com/rackspace/salus/telemetry/model/Resource.java
```diff
@@ -18,20 +18,32 @@
...
-import javax.validation.constraints.NotNull;
-import java.io.Serializable;
-import java.util.Map;
 
 @Entity
 @Table(name = "resources",
-        uniqueConstraints={@UniqueConstraint(columnNames={"tenant_id","resource_id"})})
+    uniqueConstraints = {
+        // also enables indexed query by tenant ID
+        @UniqueConstraint(columnNames = {"tenant_id", "resource_id"})
```
harageth: Is this doing individual indexes on each column or compound indexes?

itzg: It's a compound index

image

Sorry my comment confused things there.

harageth: I need to refresh some of my compound index knowledge but since the order matters I dont think it will use this index if a tenant_id is not supplied but a resource_id is supplied. Since the tenant_id portion is the first part of the index it would be the equivalent of doing a string search for "*restOfTheString"

itzg: Here's some docs about it https://dev.mysql.com/doc/refman/8.0/en/multiple-column-indexes.html

src/main/java/com/rackspace/salus/telemetry/model/Resource.java
```diff
@@ -18,34 +18,61 @@
...
     @ElementCollection(fetch = FetchType.EAGER)
     @CollectionTable(name="resource_labels", joinColumns = @JoinColumn(name="id"))
     Map<String,String> labels;
 
+    /**
+     * Unlike labels, metadata is not indexed and not used for resource/monitor matching.
+     */
+    @ElementCollection(fetch = FetchType.EAGER)
+    @CollectionTable(name="resource_metadata")
+    Map<String,String> metadata;
```
itzg: @harageth at first it seemed easy adding this metadata field; however, it quickly got complicated adding the additional JOIN in `ResourceManagement.getResourcesFromLabels`. Since these values are not used for lookups I'm thinking I'll switch this with a CLOB column that stores a JSON object.

harageth: Seems reasonable to me since they aren't being used for lookups. Makes it a lot easier to serialize/deserialize

PR #43, opened 2019-04-25
Add initial zone events

itzg:
# Resolves

Part of https://jira.rax.io/browse/SALUS-263

# What

Adds the initial zone events. There will be a corresponding PR for the new zone management code.

## How to test

A unit test was added for polymorphic parsing of zone events.


src/main/java/com/rackspace/salus/telemetry/model/Monitor.java
```diff
@@ -69,6 +69,6 @@
     @Enumerated(EnumType.STRING)
     ConfigSelectorScope selectorScope;
 
-    @ElementCollection
+    @ElementCollection(fetch = FetchType.EAGER)
```
harageth: Is there a caching thing you are trying to avoid here?

itzg: No, it's because by default JPA will lazy load relational fields like this. It'll intercept the field access and later load the association, but some reason it said the JPA session was closed at this point

https://github.com/racker/salus-telemetry-monitor-management/blob/679f6a992d46d5b57c842bbf41686a8110000384/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java#L412

src/main/java/com/rackspace/salus/telemetry/messaging/ZoneEnvoyOfResourceChangedEvent.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package com.rackspace.salus.telemetry.messaging;
+
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+@EqualsAndHashCode(callSuper = true)
+@Data
+public class ZoneEnvoyOfResourceChangedEvent extends ZoneEvent {
```
jjbuchan: I've read this about 10 times and am still confused.

_continues reading PRs_

src/main/java/com/rackspace/salus/telemetry/model/Monitor.java
jjbuchan: Yeah, if you don't do eager fetches you have to wrap everything in the same transaction... or something like that.  I hit this in the first sql prs and think i went with the eager option too.

itzg:
> I'll circle back around to this after the others, but the class names don't quite flow off the tongue.

Agreed. This was after a previous round of my own renaming and I still don't like them.

One indicates that an Envoy with its resource ID has newly registered in a given zone.

The other indicates that an Envoy momentarily went down and re-attached with the same resource ID, but with a new envoy ID.

src/main/java/com/rackspace/salus/telemetry/model/Monitor.java
itzg: I made some quick attempts at `@Transactional` wrapping, but it didn't help. I convinced myself that in that section of code, that the zones always need to be evaluated.

itzg:
@jjbuchan maybe they should be
- NewEnvoyZoneEvent or NewResourceZoneEvent
- ReattachedEnvoyZoneEvent or ReattachedResourceZoneEvent

jjbuchan:
That sounds much better.  You might wanna wait until I finish reviewing before spending time changing it though.

PR #44, opened 2019-05-01
Support conversion from Spring Data Page into PagedContent

itzg:
# What

The excessiveness of the default serialization of Spring Data's `Page` finally got on my nerves seeing content like:

```json
{
  "content": [
    {
      "monitorId": "9542a7d0-7813-429b-bfee-474e02598d59",
      "zoneTenantId": "",
      "zoneId": "public/dev",
      "resourceTenant": "aaaaaa",
      "resourceId": "test1",
      "agentType": "TELEGRAF",
      "renderedContent": "{\"type\":\"ping\",\"urls\":[\"localhost\"]}",
      "envoyId": "1844b7c8-6c17-11e9-96c9-6a00027f65d0"
    }
  ],
  "pageable": {
    "sort": {
      "unsorted": true,
      "sorted": false,
      "empty": true
    },
    "offset": 0,
    "pageSize": 20,
    "pageNumber": 0,
    "unpaged": false,
    "paged": true
  },
  "totalElements": 1,
  "last": true,
  "totalPages": 1,
  "size": 20,
  "number": 0,
  "sort": {
    "unsorted": true,
    "sorted": false,
    "empty": true
  },
  "numberOfElements": 1,
  "first": true,
  "empty": false
}
```

# How

The new `fromPage` prepares us to be able to convert all of our REST APIs that current return a `Page` with an easy drop-in conversion to `PagedContent`.

With this response type conversion, the same response above becomes much cleaner as:

```json
{
  "content": [
    {
      "monitorId": "9542a7d0-7813-429b-bfee-474e02598d59",
      "zoneTenantId": "",
      "zoneId": "public/dev",
      "resourceTenant": "aaaaaa",
      "resourceId": "test1",
      "agentType": "TELEGRAF",
      "renderedContent": "{\"type\":\"ping\",\"urls\":[\"localhost\"]}",
      "envoyId": "1844b7c8-6c17-11e9-96c9-6a00027f65d0"
    }
  ],
  "number": 0,
  "totalPages": 1,
  "totalElements": 0,
  "last": true,
  "first": true
}
```

## How to test

A Unit test was added


itzg:
> It'll probably break some tests in other repos. I think I tested a variety of the size/pages fields since I wasn't entirely sure what we'd be using.

This changes itself probably won't, but yes, as we swap out response types the unit tests will likely need to be updated. The structure of the class is an exact subset of `Page`, so is somewhat JSON compatible with `Page`.

PR #46, opened 2019-05-06
Rename selector scope ALL_OF and use underscore delimiter and remove MonitorEvent

itzg:
# Resolves

Part of
- https://jira.rax.io/browse/SALUS-313
- https://jira.rax.io/browse/SALUS-314
- https://jira.rax.io/browse/SALUS-315

# What

Finally rename that "ALL_OF" enumeration value to directly convey what it always meant: the monitor was a local/agent one vs the remote monitors.

While further testing templated label retrieval, it was discovered that dot-delimited labels will cause usability issues:

https://github.com/racker/salus-telemetry-monitor-management/pull/27/files#diff-2332b138120e6010ec20463e65fd4465R129

Part of this PR changes the label namespace delimiter to underscore.

Also removes `MonitorEvent` since https://github.com/racker/salus-telemetry-monitor-management/pull/27 removed all remaining use of that event.

## How to test

N/A

# TODO

There will be a monitor management PR that will include the code affected by this rename.


src/main/java/com/rackspace/salus/telemetry/model/ValidLabelKeysValidator.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.regex.Pattern;
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+
+/**
+ * Validates that the annotated map contains that keys that start with a letter and then contain
+ * only letters, digits, and underscore.
+ */
+public class ValidLabelKeysValidator implements ConstraintValidator<ValidLabelKeys, Map<String,String>> {
+   private static final Pattern VALID_KEY = Pattern.compile("[A-Za-z][A-Za-z0-9_]*");
```
jjbuchan:
```suggestion
   private static final Pattern VALID_KEY = Pattern.compile("[A-Za-z0-9_]+");
```

Am I being dumb, or should this not be what's used?

jjbuchan: Some test cases.

`1_dfw`
`_this_is_a_test`

```
```
 empty string (although, probably fails elsewhere in our code)

itzg: Your suggested change goes against the javadoc I stated twice:
> keys that start with a letter and then contain only letters, digits, and underscore

but with that said, I don't feel strongly prohibiting the label keys starting with numbers and underscores. I was just following suit of typical programming identifier naming rules.

itzg: I'll add those as negative test cases...unless you want the leading digit/underscore rule.

jjbuchan: > I stated twice:

Oh.  My mistake.  I definitely must have skimmed over that.

But I was basing the comment on the validation message that's returned. `All label names must consistent of alpha-numeric or underscore`.  Maybe that needs to be adapted to be clearer.

jjbuchan: I guess I was thinking there's also no reason to disallow them from starting with those characters?

itzg: Ah yes, I was avoiding a wordy error message. So I like the idea of keeping the pattern simple.

jjbuchan:
Ooh, conflicts.

itzg:
Should be easy conflicts to resolve since the class went away :)

itzg:
...actually, I transplanting your change @jjbuchan over to the moved one:

```
    @CollectionTable(name="monitor_zones", joinColumns = @JoinColumn(name="monitor_id"))
```

PR #47, opened 2019-05-09
Better define zones field in a monitor

jjbuchan:
I added this while debugging another problem, and although it might not be required, I think it's better to have it than not so that we clearly set the column name.

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/model/Monitor.java
+++ b/src/main/java/com/rackspace/salus/telemetry/model/Monitor.java
@@ -50,7 +50,7 @@ public class Monitor implements Serializable {
     String monitorName;
 
     @ElementCollection(fetch = FetchType.EAGER)
-    @CollectionTable(name="monitor_label_selectors", joinColumns = @JoinColumn(name="id"))
+    @CollectionTable(name="monitor_label_selectors", joinColumns = @JoinColumn(name="monitor_id"))
     Map<String,String> labelSelector;
 
     @NotBlank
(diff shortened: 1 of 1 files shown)


src/main/java/com/rackspace/salus/telemetry/model/Monitor.java
```diff
@@ -70,5 +70,6 @@
     ConfigSelectorScope selectorScope;
 
     @ElementCollection(fetch = FetchType.EAGER)
+    @CollectionTable(name="monitor_zones", joinColumns = @JoinColumn(name="id"))
```
itzg: The default `monitor_id` seemed ok to me and the table was already being named `monitor_zones`. A column name `id` makes me think it means zone ID.

jjbuchan: I agree.  This was a copy paste and I didn't think to change the join column.  I've changed it for the two entries in this file.

I left it for `@CollectionTable(name="resource_labels", joinColumns = @JoinColumn(name="id"))` because there isn't an obvious better value there.

> the table was already being named monitor_zones

This is the same as all our columns though.  I expect if we removed all the `@Column` name values pretty much everything would still work... but it also might not.  This just makes it less ambiguous.

jjbuchan:
> Can you clarify how this change helped debugging?

It didn't.  I added this as part of the repository `Contains` problem where that option worked in a unit test but not in a more manual test.  However, I thought it was better to have it than not.

PR #52, opened 2019-05-21
Add reattached-envoy resource event

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-360

# What

Adds a new resource event that will be produced when resource management detects an Envoy is re-attaching to a resource that previously had an Envoy.

Design background is provided in https://github.com/racker/salus-telemetry-bundle/pull/50

## How to test

Unit tests added and updated


src/main/java/com/rackspace/salus/telemetry/messaging/ReattachedEnvoyResourceEvent.java
```diff
@@ -0,0 +1,34 @@
...
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+/**
+ * Produced by Resource Management, this event indicates that a new Envoy instance has re-attached
+ * with a pre-existing resource ID. Local monitors that were previously bound to the resource
+ * should be re-bound to the Envoy given in this event.
+ */
+@Data @EqualsAndHashCode(callSuper = true)
+@KafkaMessageKey(properties = {"tenantId", "resourceId"})
```
jjbuchan: I forgot about this annotation.  I don't think we're using it for the remote monitor events.

itzg: I had forgotten about it too and thought I'd revive it here.

itzg:
@jjbuchan ready for re-review

src/main/java/com/rackspace/salus/telemetry/messaging/ResourceEvent.java
```diff
@@ -16,19 +16,32 @@
...
+    /**
+     * Indicates that the labels or metadata of the resource have changed.
+     * Label selection should and content rendering should be re-evaluated.
+     */
+    boolean labelsChanged;
+
+    /**
+     * Indicates that the resource was deleted.
+     */
+    boolean deleted;
```
jjbuchan: My only question is on the purpose of this field, which I already commented on in the mon-mgmt pr.

itzg: As mentioned over there, it's for completeness and informational/troubleshooting purposes. With this, the event conveys all the reasons for a resource change to be produced:

- labels/metadata changed
- envoy re-attached to the resource
- both of the above
- resource deleted

PR #58, opened 2019-07-16
Support API to get label namespaces

itzg:
# Resolves

Part of https://jira.rax.io/browse/SALUS-450

# What

Provides a way for a Salus REST API to tell external users what label namespaces (i.e. prefixes) are known by the system. That provides a hint not to use labels with those prefixes prior to attempting an API call (like resource update) and having it fail.


src/main/java/com/rackspace/salus/telemetry/model/LabelNamespaces.java
```diff
@@ -75,5 +77,9 @@ public static boolean labelHasNamespace(String label, String namespace) {
     return label.startsWith(namespace + DELIM);
   }
 
+  public static Collection<String> getNamespaces() {
+    return Collections.unmodifiableCollection(ourNamespaces);
```
jjbuchan: So this is essentially a safety guard to prevent someone from trying to modify this in another service?

i.e. we should be fine with or without it, but it's just "safer" to add it?

itzg: Yeah, since the namespaces are supposed to be a "constant" I wanted to guard the return value...however, your question made me realize I can change the declaration in the first place to make `ourNamespaces` unmodifiable.

src/main/java/com/rackspace/salus/telemetry/model/LabelNamespaces.java
```diff
@@ -39,15 +43,13 @@
    */
   public static final String MONITORING_SYSTEM_METADATA = "system_metadata";
 
-  private static final Set<String> ourNamespaces = new HashSet<>();
+  private static final Set<String> ourNamespaces = unmodifiableSet(new HashSet<>(asList(
```
jjbuchan: I think you could just do `Set.of(...)`.  I had read about that the other day when looking at the difference between unmodifiable vs. immutable.

...although maybe that depends on bumping our Java version.  Should we start doing that since I keep seeing more use-cases?

itzg: Yeah, doesn't seem to be in Java 8 and yes we really should upgrade to Java 11. The harder part was coordinating our docker salus-app-base declaration. It looks like the 11 images are now more viable

https://github.com/docker-library/docs/blob/master/openjdk/README.md#supported-tags-and-respective-dockerfile-links

Before when I looked there were only Oracle choices there.

PR #60, opened 2019-07-30
Add policy events

jjbuchan:
Adds the kafka messages that will be used for policy events.

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/messaging/PolicyEvent.java
+++ b/src/main/java/com/rackspace/salus/telemetry/messaging/PolicyEvent.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2019 Rackspace US, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.rackspace.salus.telemetry.messaging;
+
+import com.rackspace.salus.common.messaging.KafkaMessageKey;
+import java.util.UUID;
+import lombok.Data;
+
+@Data
+@KafkaMessageKey(properties = {"policyId", "tenantId"})
+public abstract class PolicyEvent {
+  UUID policyId;
+  String tenantId;
+}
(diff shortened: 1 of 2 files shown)


src/main/java/com/rackspace/salus/telemetry/messaging/PolicyEvent.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package com.rackspace.salus.telemetry.messaging;
+
+import com.rackspace.salus.common.messaging.KafkaMessageKey;
+import java.util.UUID;
+import lombok.Data;
+
+@Data
+@KafkaMessageKey(properties = {"policyId", "tenantId"})
```
GeorgeJahad: Probably doesn't matter,  but don't you just want this annotation in the non abstract classes?  Does this accomplish anything here?

jjbuchan: Originally I only added it here hoping the other classes would inherit it, which never ended up being the case.

In the end, I left it since it felt safer in case it ever ended up being needed vs. it being needed and us forgetting to add it back.

I'm happy to remove it though if anyone wants that.

GeorgeJahad: It's fine by me, but the professor might disagree.

itzg: This call to `getAnnotation` 

https://github.com/racker/salus-common/blob/484614b7ecc2521a239bc25c02c7aa34058f6a51/src/main/java/com/rackspace/salus/common/messaging/KafkaMessageKeyBuilder.java#L33

seems like it should grab annotations from the inherited classes, but the javadocs aren't very clear about that. I'm guessing that `getAnnotation` is broader than `getDeclaredAnnotation` by looking at its javadocs:

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Class.html#getDeclaredAnnotation(java.lang.Class)

jjbuchan: I didn't look as far as you, but running the service would complain that the key wasn't set if I didn't add it to the Monitor Policy Event.

itzg: Even if you're not using it via `KafkaMessageKeyBuilder` the annotation is still useful as a means to document the intended partitioning. So, I'd say leave it, but probably not needed on the `MonitorPolicyEvent`.

itzg: How did it complain exactly? Since it's an annotation we invented, I'm surprised anything would care.

jjbuchan: Because I call `buildMessageKey` in the PolicyEventProducer.

itzg: OK :) that'll do it then. I guess the speculation about `getAnnotation` doesn't hold then. I wonder why they even provide `getDeclaredAnnotation` then.

PR #62, opened 2019-08-07
Add entities/repositories from other services

jjbuchan:
This can be reviewed one commit at a time.

# What

Used in combination with:

https://github.com/racker/salus-policy-management/pull/4
https://github.com/racker/salus-telemetry-monitor-management/pull/68
https://github.com/racker/salus-telemetry-resource-management/pull/48
https://github.com/racker/salus-telemetry-agent-catalog-management/pull/5

We moved the entities/repositories from those repos to here so they can be shared across services.

One thing I had to do was duplicate the `ResolvedZone.PUBLIC` field we use elsewhere within the `Zone` object so we didnt have to make `model` depend on `etcd-adapter`.  we may want to refactor that field out of etcd-adapter later on.

## How to test

Run tests.

# Why

Rather than having our services perform api calls to communicate with eachother they will just read from the shared database. This lowers the risk of hitting circular dependency problems and should reduce latency.

# TODO

Do all the other services.

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/entities/MonitorPolicy.java
+++ b/src/main/java/com/rackspace/salus/telemetry/entities/MonitorPolicy.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2019 Rackspace US, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.rackspace.salus.telemetry.entities;
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.Table;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import org.hibernate.validator.constraints.NotBlank;
+
+@Entity
+@Table(name = "policy_monitors")
+@Data
+@EqualsAndHashCode(callSuper = true)
+public class MonitorPolicy extends Policy {
+
+  @NotBlank
+  @Column(name="name")
+  String name;
+
+  @NotBlank
+  @Column(name="monitor_id")
+  String monitorId;
+}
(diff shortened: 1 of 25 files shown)


GeorgeJahad:
As a suggestion for next time, when a file is copied into a repository, I prefer the original commit to contain the original unaltered file.  That way I can do a diff in place.

src/main/java/com/rackspace/salus/telemetry/repositories/AgentInstallRepository.java
```diff
@@ -0,0 +1,34 @@
...
+
+import com.rackspace.salus.telemetry.entities.AgentInstall;
+import java.util.List;
+import java.util.Optional;
+import java.util.UUID;
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.repository.CrudRepository;
+
+public interface AgentInstallRepository extends CrudRepository<AgentInstall, UUID> {
```
GeorgeJahad: It looks like this was changed from a JpaRepository?  was that intentional?

jjbuchan: It wasn't changed - https://github.com/racker/salus-telemetry-agent-catalog-management/blob/master/src/main/java/com/rackspace/salus/acm/repositories/AgentInstallRepository.java#L27

jjbuchan:
I changed pretty much as little as possible to get things to work in the different repos.

What I could have done is make PRs in the individual repos to get them to work in a similar manner, then once those are approved rip the files out and move them over as is.

src/main/java/com/rackspace/salus/telemetry/repositories/AgentInstallRepository.java
GeorgeJahad: sorry, my bad.

GeorgeJahad:
Yes, I can see you've changed as little as possible.  (And I appreciate that.)  But what I was trying to get at, is that if the first commit has the unaltered file, immediately after it is copied over, then we actually have a history of the changes, and it makes it easier to diff.  

I know the package statement will be wrong and the file won't compile because it is in a new directory, but I still like to see it in the history, before the fixes are made.  Otherwise, it's like there is a gap in the history.  Does that make sense?

jjbuchan:
Ah yeah, I see what you mean.  I was trying to make the commits as meaningful as possible, but yeah, had I done two for each step that'd be better and avoid the extra step in the option I noted.

src/main/java/com/rackspace/salus/telemetry/entities/MonitorPolicy.java
```diff
@@ -0,0 +1,39 @@
...
+@EqualsAndHashCode(callSuper = true)
+public class MonitorPolicy extends Policy {
+
+  @NotBlank
+  @Column(name="name")
+  String name;
+
+  @NotBlank
+  @Column(name="monitor_id")
+  String monitorId;
```
itzg: Something for a future PR: since these are all now guaranteed to be in the same database we could even make these "soft" foreign key relationships be real JPA/FK relationships. For example, I think this one would be a `@OneToOne`.

PR #63, opened 2019-08-09
Update entities to work with policy management

jjbuchan:
# What

Updates some things which will be needed to complete policy management functionality.

# How

 * Adds column names to entities that didn't have it to make index definitions more clear
 * Add tenantId to boundMonitor
   * On the DTOs this will replace resourceTenant
 * Moved POLICY_TENANT to the entity so it can be shared across repos.
 * Changed monitorId to a UUID to be consistent with the id field of the Monitor entity.
 * Adds some new repository methods
 * Adds PolicyMonitorUpdateEvent
 * Adds TenantChangeEvent

# Why

Bound Monitors must now have tenantId in their primary key since policy monitors will be shared across all tenants.  The previous key is no longer guaranteed to be unique if multiple tenants have the same resource ids.

PolicyMonitorUpdateEvent will be consumed by both policy mgmt and mon mgmt when a policy monitor is updated.  Policy mgmt will receive the event with a null tenantId, then determine which tenants it is relevant to and send out individual event for each of them which mon mgmt will consume and handle.

TenantChangeEvent will be consumed by mon mgmt to ensure all policy monitors are applied correctly.  e.g. the tenant's account type could be updated which should trigger a refresh of what policies are applied to it.

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/entities/Monitor.java
+++ b/src/main/java/com/rackspace/salus/telemetry/entities/Monitor.java
@@ -45,8 +45,7 @@ import org.hibernate.validator.constraints.NotBlank;
 
 @Entity
 @Table(name = "monitors", indexes = {
-    @Index(name = "monitors_by_tenant", columnList = "tenant_id"),
-    @Index(name = "by_tenant_and_resource", columnList = "tenant_id, resource_id")
+    @Index(name = "monitors_by_tenant_and_resource", columnList = "tenant_id, resource_id")
 })
 @NamedQueries({
     @NamedQuery(name = "Monitor.getDistinctLabelSelectors",
--- a/src/main/java/com/rackspace/salus/telemetry/entities/Monitor.java
+++ b/src/main/java/com/rackspace/salus/telemetry/entities/Monitor.java
@@ -53,6 +53,9 @@ import org.hibernate.validator.constraints.NotBlank;
 })
 @Data
 public class Monitor implements Serializable {
+
+    public static final String POLICY_TENANT = "_POLICY_";
+
     @Id
     @GeneratedValue
     @org.hibernate.annotations.Type(type="uuid-char")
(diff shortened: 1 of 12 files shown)


src/main/java/com/rackspace/salus/telemetry/entities/Monitor.java
```diff
@@ -53,6 +53,9 @@
 })
 @Data
 public class Monitor implements Serializable {
```
itzg: Just making a note that the index name `by_tenant` apparently needs to be named more uniquely, like `monitors_by_tenant` since there's other indices now in the same context with the same name.

src/main/java/com/rackspace/salus/telemetry/messaging/TenantChangeEvent.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package com.rackspace.salus.telemetry.messaging;
+
+import com.rackspace.salus.common.messaging.KafkaMessageKey;
+import lombok.Data;
+
+@Data
+@KafkaMessageKey(properties = {"tenantId"})
+public class TenantChangeEvent {
```
itzg: I wonder if this needs a javadoc if only because outside the using-context I'm amused by how vague/broad this event sounds.

itzg: and/or maybe this class name should mention "policy", such as `TenantPolicyChangeEvent`

jjbuchan: I keep struggling wit the right name for the tenant stuff.  I've pushed a change to this now and am thinking later I should change TenantMetadata to TenantPolicyInfo or something like that.

src/main/java/com/rackspace/salus/telemetry/entities/Monitor.java
```diff
@@ -45,7 +45,8 @@
 
 @Entity
 @Table(name = "monitors", indexes = {
-    @Index(name = "monitors_by_tenant", columnList = "tenant_id")
+    @Index(name = "monitors_by_tenant", columnList = "tenant_id"),
+    @Index(name = "by_tenant_and_resource", columnList = "tenant_id, resource_id")
```
itzg: You can actually remove the `monitors_by_tenant` index since btree indices can be satisfied by partial, left-most columns. While you're in here, this should be named `monitors_by_tenant_resource` to avoid index name conflicts.

PR #64, opened 2019-08-15
added resource id to monitor

GeorgeJahad:
# Resolves

related to https://jira.rax.io/browse/SALUS-494

# What

Adds model entries for monitor resourceId


src/main/java/com/rackspace/salus/telemetry/repositories/MonitorRepository.java
```diff
@@ -27,6 +27,8 @@
 
     Page<Monitor> findByTenantId(String tenantId, Pageable pageable);
 
+    Monitor findByTenantIdAndResourceId(String tenantId, String resourceId);
```
itzg: It seems like this should return `List<Monitor>` since the user could create several resource-selecting monitors.

GeorgeJahad: Doh!  excellent point!

PR #72, opened 2019-08-29
Allowing monitoring interval to be configurable per monitor

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-592

# What

Adds monitoring interval column to monitor entity

## How to test

Gets tested in monitor-management


itzg:
> Thanks for the comment about the SQL INTERVAL.

When I first tried naming the column just "interval" it gave a strange, misleading schema error, so didn't want to wander into that again.

harageth:
> > Thanks for the comment about the SQL INTERVAL.
> 
> When I first tried naming the column just "interval" it gave a strange, misleading schema error, so didn't want to wander into that again.

And it looks like its actually built into the query language actually. So you can use INTERVAL in a query against a timestamp if you really wanted. But yeah those kinds of "schema errors" are really weird since its not really part of the user defined schema and is really just an implicit function call.

PR #75, opened 2019-09-17
add db plugin monitor types

GeorgeJahad:


GeorgeJahad:
I thoughtlessly checked in the monitor management pr, before this one.  so i need to check this one in now cause the build is broken without it.  I'll revert both if there is a problem with that.

PR #76, opened 2019-09-27
Added entity, repository, and translator structure

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-481

# What

Per this discussion

https://github.com/racker/salus-telemetry-monitor-management/pull/90#pullrequestreview-293857720

the entity, repository, and monitor translator classes are added here in salus-model to follow the existing, centralized pattern of entities and repositories.

**NOTE** we'll need to keep on eye on the maintenance churn of the translators and might need to think of a strategy to refactor those. 

## How to test

Unit test moved over

# Related PRs

- https://github.com/racker/salus-telemetry-monitor-management/pull/90


src/test/java/com/rackspace/salus/telemetry/translators/RenameFieldTranslatorTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+    final RenameFieldTranslator translator = new RenameFieldTranslator()
+        .setFrom("field-match")
+        .setTo("new-field");
+
+    final ObjectNode contentTree = objectMapper.valueToTree(content);
+
+    translator.translate(contentTree);
+
+    assertThat(contentTree).hasSize(2);
```
itzg: @GeorgeJahad here's the fix for your previous comment https://github.com/racker/salus-telemetry-monitor-management/pull/90#pullrequestreview-294071540

itzg:
> I didn't check this extensively, but it looks the same as the files I already approved, so I'm approving it again. Let me know if there are any significant diffs, (other than the size check.)

You're correct, that's the only diff.

PR #78, opened 2019-10-09
Add model changes to the AgentInstall to have a LabelSelectorMethod

harageth:
# Resolves

JIRA-606

# What

Add the LabelSelectorMethod to the AgentInstall storage model

## How to test

Compile this and agent-catalog management on my branch over there then create a new AgentInstall object.

# Why

This was how we stored the same information on Monitors

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/entities/AgentInstall.java
+++ b/src/main/java/com/rackspace/salus/telemetry/entities/AgentInstall.java
@@ -37,6 +37,7 @@ import javax.persistence.NamedQuery;
 import javax.persistence.Table;
 import javax.validation.constraints.NotNull;
 import lombok.Data;
+import org.hibernate.annotations.ColumnDefault;
 import org.hibernate.annotations.CreationTimestamp;
 import org.hibernate.annotations.UpdateTimestamp;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/rackspace/salus/telemetry/entities/AgentInstall.java
```diff
@@ -70,6 +74,11 @@
   @CollectionTable(name="agent_install_label_selectors", joinColumns = @JoinColumn(name="agent_install_id"))
   Map<String,String> labelSelector;
 
+  @NotNull
+  @Enumerated(EnumType.STRING)
+  @Column(name="label_selector_method")
+  LabelSelectorMethod labelSelectorMethod = LabelSelectorMethod.AND;
```
itzg: Since this column was messy to retrofit in the existing monitor records in the dev cluster, I'm curious if adding this Hibernate annotation I just discovered would help until we have formal schema management in our system:

```
@ColumnDefault("AND")
```

harageth: Oh yeah that seems like it will make the migration go a lot better. I'll make that change

PR #80, opened 2019-10-15
Add monitor translator for joining host and port fields

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-483

# What

For plugins like `net_response` it will be better to have our users provide the host and port as two distinct fields, yet provide telegraf with the joined "host:port" field.

# How

Added `joinHostPort` translator.

## How to test

Added unit tests


src/test/java/com/rackspace/salus/telemetry/translators/JoinHostPortTranslatorTest.java
```diff
@@ -0,0 +1,84 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import java.io.IOException;
+import org.junit.Test;
+
+public class JoinHostPortTranslatorTest {
+  private final ObjectMapper objectMapper = new ObjectMapper();
+
+  @Test
+  public void testTranslate_typical() throws IOException {
```
GeorgeJahad: I had a little trouble seeing what the difference was between these two tests was.  Maybe if this was called "testTranslate_numericPort()"  it would have been easier to see.

itzg: Good idea.

itzg:
> My only thought was whether we should have some negative test cases, but I'm assuming any validation on those fields would happen prior to the translator so those are irrelevant.

Yeah, with the simplistic nature of the translators I struggled to come up with some negative test cases other than`testTranslate_missingPort`. What were some ideas you had for those?

PR #83, opened 2019-11-04
Add a unique name field to MonitorTranslationOperator

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-671

# What

For the data loader mechanism we need a stable and unique way to refer to monitor translation operators that have already been loaded.

# How

Add a `name` field which is declared with a uniqueness constraint.

# TODO

Will also need to update the DTO in Monitor Management


src/test/java/com/rackspace/salus/telemetry/entities/MonitorTranslatorJsonStringTypeTest.java
```diff
@@ -49,6 +49,7 @@
   @Test
   public void testPersistRoundtrip() {
     final MonitorTranslationOperator operator = new MonitorTranslationOperator()
+        .setName(("cpu-rename"))
```
jjbuchan: double `(( )) `?

itzg: Oops, I've noticed auto complete trying to do that before. I wonder how that compiled for me...fixing it.

PR #84, opened 2019-11-26
aded ORACLE as an agent type and the new monitor types

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-693

# What

Adds the Oracle Agent and new oracle agent monitor types

# How

It follows our paradigm for adding new monitors to MonitorType and adding the Oracle Agent to AgentType

## How to test

Run the unit tests in the branch named the same thing in MonitorManagement

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/model/MonitorType.java
+++ b/src/main/java/com/rackspace/salus/telemetry/model/MonitorType.java
@@ -32,7 +32,7 @@ public enum MonitorType {
   sqlserver,
   log,
   system,
-  tablespace,
-  dataguard,
-  rman
+  oracle_tablespace,
+  oracle_dataguard,
+  oracle_rman
 }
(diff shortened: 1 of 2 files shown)


src/main/java/com/rackspace/salus/telemetry/model/MonitorType.java
```diff
@@ -31,5 +31,8 @@
   postgresql,
   sqlserver,
   log,
-  system
+  system,
+  tablespace,
```
itzg: I missed the enum field in the other PR when suggesting the "oracle" qualifier. Since you haven't merged that one yet, can you go ahead and qualify these. It's not quite proper Java convention, but could also name these like "oracle_tablespace".

harageth: Yeah I will switch them to the same name. Only caught that I hadn't merged this in because of the build failure on the other one, so I figured I would give you a chance to suggest that the names remain the same. imo it is a little weird qualifying them like this but not enough that I care more than this comment.

PR #85, opened 2019-12-03
Add flyway for schema management

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-684
https://jira.rax.io/browse/SALUS-574

# What

Implements the model repo changes as proposed in https://github.com/Rackspace-Segment-Support/salus-docs-internal/pull/19

# How

- Started with a schema dump from my local MySQL after Hibernate had applied the latest schema (https://github.com/racker/salus-telemetry-model/commit/8be29b4092790c71121b426f8b831fd030055182)
- Split that out into vendor specific files for the hibernate sequence and did some cleanup of foreign key and constraint declarations (https://github.com/racker/salus-telemetry-model/commit/328e387a616e8c35d68dd349382f32366802adf3)
- Added a helper annotation `@EnableSalusJpa` to activate flyway, set hibernate auto DDL to validate, and encapsulate the repo/entity scanning we usually declare (https://github.com/racker/salus-telemetry-model/pull/85/commits/6f6228beb49c1a3675650f93c77da3b11957b5ef)
- This commit to the unit tests shows how that new annotation gets used (https://github.com/racker/salus-telemetry-model/commit/89036364a8026ecf638ef0d327c7e3252389f5f7)
- hibernate-types-52 wasn't cooperating in some cases, so generalized the existing JSON custom hibernate type to allow for any object type (https://github.com/racker/salus-telemetry-model/commit/8fedf546f7652fac584ad10f99d54e9199931fb9 and https://github.com/racker/salus-telemetry-model/commit/bd4ea795b652e711b1ada591c44f17a94661be8d)
- while I was in here, brought over the remaining entities that weren't yet in this repo (https://github.com/racker/salus-telemetry-model/commit/f7106ad8f9f9b7a9d4a2b225652da69d12a33881)

## How to test

Unit tests and started a few of our applications with mysql container's volume removed and rebuilt from scratch with flyway.

# TODO

There will be smaller PRs for each app that uses JPA to switch over to `@EnableSalusJpa`.


src/main/resources/spring-flyway.properties
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# Rather than the default classpath:db/migration use vendor specific organization to allow
+# for our embedded H2 in unit testing and MySQL otherwise
+spring.flyway.locations = classpath:db/migration/common,classpath:db/migration/{vendor}
```
GeorgeJahad: this puzzled me a bit.  a reference to would be nice: https://docs.spring.io/spring-boot/docs/current/reference/html/howto.html#howto-execute-flyway-database-migrations-on-startup

src/main/java/com/rackspace/salus/telemetry/entities/JsonClobType.java
```diff
@@ -0,0 +1,119 @@
...
+ * <p>
+ *   Field to be encoded as JSON in a CLOB SQL column should specify "json" as the
+ *   {@link Type#type}.
+ * </p>
+ * <p>
+ * <em>NOTE</em> the hibernate-types library wasn't a viable solution since the SQL type
+ * declared by the JSON string types is hardcoded as "OTHER".
+ * </p>
+ */
+public class JsonClobType
```
GeorgeJahad: I understand the intent of this class, but it's still pretty hard for me to follow.  Do you have a favorite tutorial explaining what you are doing here?

src/main/resources/spring-flyway.properties
itzg: Excellent point. I'll include that link since that's exactly the section I used to discover the option.

src/main/java/com/rackspace/salus/telemetry/entities/JsonClobType.java
itzg: Yeah, this custom Hibernate type stuff is horrible looking. And yes, I have an article I used for this re-implementation, so I'll include that in the docs.

itzg:
For our existing deployments we'll use [flyway's baseline command](https://flywaydb.org/documentation/commandline/baseline) to first baseline at V1.1 of the migrations.

Locally I did that with
```
docker run --rm --network=telemetry-infra_default \
  flyway/flyway \
  -url=jdbc:mysql://telemetry-infra_mysql_1:3306/default \
  -user=dev -password=pass \
  baseline -baselineVersion=1.1 -baselineDescription="hibernate sequence"
```

Then the first application deployed to our cluster will apply the V1.2 migration (https://github.com/racker/salus-telemetry-model/pull/85/commits/3def190edf551ee96921608e489ec7edff45f4fb) which fixes up the "JSON column" types to be `longtext` rather than the mixture of column types we had before.

itzg:
@jjbuchan @GeorgeJahad @harageth I believe the command that should work for you (did for me at least), is to baseline your mysql "infra" container using:

```
docker run --rm --network=telemetry-infra_default flyway/flyway -url=jdbc:mysql://telemetry-infra_mysql_1:3306/default -user=dev -password=pass baseline -baselineVersion=1.1 -baselineDescription="hibernate sequence"
```

Worst case if that doesn't work, you can stop mysql, remove the container, remove its volume, and let it start fresh:

```
docker-compose stop mysql
docker-compose rm mysql
docker volume rm telemetry-infra_mysql
docker-compose up -d mysql
```

PR #88, opened 2019-12-12
Switching to maven's version parsing

itzg:
# Resolves

Part of https://jira.rax.io/browse/SALUS-724

# What

The java-semver library is too strict with its version parsing, since it only handles compliant semantic versions.

# How

Switch over to the version parsing included in [maven's artifact library](https://maven.apache.org/ref/3.3.3/maven-artifact/apidocs/org/apache/maven/artifact/versioning/package-frame.html); however, the model repo doesn't need either one of these libraries declared. The apps should declare the dependency instead.


jjbuchan:
Ah, I see you said

> Even though we'll eventually switch the Oracle Agent to use semantic versioning, we should probably change that code to be more tolerant of "non-compliant" versions

itzg:
> No real opinion but is this the right choice or should the agent version just change to 0.1.0?

That was the same question I was pondering, but I'm thinking down the road there might be a third party agent we use that uses some kind of non-semver scheme. I thought it would be better to switch version range syntax now before we have monitor translations in use.

PR #91, opened 2019-12-18
Add replace field value tranlator

jjbuchan:
Renames an existing translator to be more specific and adds another one that adds/replaces a field with a new value.

The term `replace` instead of `rename` seemed more accurate here.

The first use-case is converting the `type` field of a monitor such as `dns` to the telegraf input type `dns_query`.  Similar for the `x509` input plugin.

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/translators/ReplaceFieldValueTranslator.java
+++ b/src/main/java/com/rackspace/salus/telemetry/translators/ReplaceFieldValueTranslator.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2019 Rackspace US, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


src/main/java/com/rackspace/salus/telemetry/translators/ReplaceFieldValueTranslator.java
```diff
@@ -0,0 +1,41 @@
...
+import javax.validation.constraints.NotEmpty;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+/**
+ * Replaces a field with the new value.
+ * If the field didn't already exist it will be added.
+ */
+@Data @EqualsAndHashCode(callSuper = false)
+public class ReplaceFieldValueTranslator extends MonitorTranslator {
```
itzg: I like the choice of "replace" especially since we could also use this translator for "setting" fixed values for fields, like`query_version` in the SqlServer plugin 

...however, that begs the question, how should we do a replace for non-string values? I could see leaving this as is and later adding `ReplaceIntFieldValueTranslator`, etc.

jjbuchan: Similar thing crossed my mind and I chose to worry about that later.

I did forget that we can't really get by with `query_version` as it is now.  It will need to be removed from the POJO.

If you agree I'll add `ReplaceIntFieldValueTranslator` to this and also rename this one to `ReplaceStringFieldValueTranslator`

itzg: I'm good with that plan

PR #94, opened 2019-12-30
Add translator to rename monitor type values

jjbuchan:
Similar to the `replaceStringFieldValue` one but to improve readability since this use-case is not unique.  Discussed here - https://github.com/Rackspace-Segment-Support/salus-data-loader-content/pull/1#discussion_r362013705

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/translators/RenameTypeTranslator.java
+++ b/src/main/java/com/rackspace/salus/telemetry/translators/RenameTypeTranslator.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2019 Rackspace US, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


src/main/java/com/rackspace/salus/telemetry/translators/RenameTypeTranslator.java
```diff
@@ -0,0 +1,40 @@
...
+
+package com.rackspace.salus.telemetry.translators;
+
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import javax.validation.constraints.NotEmpty;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+/**
+ * Replaces a field with the new value.
```
jjbuchan:
```suggestion
 * Replaced the monitor type with the new value provided.
```

PR #96, opened 2020-01-06
Add exclude resource IDs field to monitor

itzg:
# Resolves

Part of https://jira.rax.io/browse/SALUS-695

# What

Adds an `excludedResourceIds` field which allows for a monitor to be selectively excluded from binding to the given resource that would be normally be selected by labels.

## How to test

Flyway/schema parts are tested by existing unit tests. Main unit testing happens via the monitor-management changes that will be in a related PR.


src/main/java/com/rackspace/salus/telemetry/entities/Monitor.java
```diff
@@ -121,6 +122,11 @@
     @NonMetadataField
     String resourceId;
 
+    @ElementCollection(fetch = FetchType.EAGER)
```
harageth: I thought it would only allow us to eagerly fetch one collection?

itzg: Yeah, I thought so too and originally didn't have this annotation. During broader unit tests I got the exception about the session being closed, added the annotation, and it actually worked. I seem to remember the issue with more than one eager fetch was during manual JPA queries.

jjbuchan: Hmm, I remember this being a problem too but now this is the 3rd on here.

itzg:
I'll go ahead and merge this so I can verify the build of https://github.com/racker/salus-telemetry-monitor-management/pull/120

PR #107, opened 2020-01-23
Add duration in seconds translator

jjbuchan:
# Resolves

https://jira.rax.io/browse/SALUS-753

# What

1. Adds a translation to convert duration strings to a raw seconds value.
1. Tiny GoDurationTranslator.java tweak
    * It was accidentally left after I wanted to set a breakpoint on that field.

# How

Simply use the `toSeconds` method.

## How to test

Run tests.

# Why

Plugins such as Ping need their values to be in seconds.

# TODO

Individual translator content still needs to be added to data loader.

### DIFF
--- a/src/test/java/com/rackspace/salus/telemetry/translators/DurationInSecondsTranslatorTest.java
+++ b/src/test/java/com/rackspace/salus/telemetry/translators/DurationInSecondsTranslatorTest.java
@@ -19,6 +19,7 @@ package com.rackspace.salus.telemetry.translators;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.NullNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import java.io.IOException;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


src/test/java/com/rackspace/salus/telemetry/translators/DurationInSecondsTranslatorTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+    final DurationInSecondsTranslator translator = new DurationInSecondsTranslator()
+        .setField("timeout");
+
+    final ObjectNode contentTree = (ObjectNode) objectMapper.readTree(content);
+
+    translator.translate(contentTree);
+
+    assertThat(contentTree).hasSize(1);
+    assertThat(contentTree.get("timeout")).isNotNull();
```
GeorgeJahad: what value should the timeout have here?

jjbuchan: I've added a new commit for this.  This is showing `timeout` exists in the payload and the new commit shows a `NullNode` will be the value.  i.e. the field exists and contains a null value.

PR #109, opened 2020-01-28
Add extra operators to kapacitor tasks

jjbuchan:
# What

Adds the ability to use more of kapacitors operators in tasks.

image

# How

Updates and enum and goes with https://github.com/racker/salus-event-engine-management/pull/33

## How to test

Added some more tests and verified manually that kapacitor accepts the new payload.

# Why

I was looking through the event engine and saw this was an easy thing to add in. It's especially useful for string metrics or integers that don't change much.

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/validators/ExpressionValidator.java
+++ b/src/main/java/com/rackspace/salus/telemetry/validators/ExpressionValidator.java
@@ -29,7 +29,8 @@ public class ExpressionValidator implements ConstraintValidator<ComparatorValida
   @Constraint(validatedBy = ExpressionValidator.class) // validator
   @Documented
   public @interface ComparatorValidation {
-    String message() default "Valid comparators are: >, >=, <, <="; // default error message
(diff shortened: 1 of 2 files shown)


src/main/java/com/rackspace/salus/telemetry/validators/ExpressionValidator.java
```diff
@@ -29,7 +29,8 @@ public boolean isValid(String comparator, ConstraintValidatorContext context) {
   @Constraint(validatedBy = ExpressionValidator.class) // validator
   @Documented
   public @interface ComparatorValidation {
-    String message() default "Valid comparators are: >, >=, <, <="; // default error message
+    String message() default "Valid comparators are: "
+        + "==, !=, >, >=, <, <=, =~, !~"; // default error message
```
jjbuchan: I wanted to change this to be dynamic based on the `convertString` map but it has to be constant in the interface which is annoying but makes sense.

PR #110, opened 2020-01-28
Updated to use testcontainers for DB testing

itzg:
# Resolves

Part of https://jira.rax.io/browse/SALUS-738

# What

The Spring Boot upgrade from 2.1.8 to 2.2.4 included an upgrade of h2 from 1.4.199 to 1.4.200. Even though that version of h2 brings some nice features like json data type support, it also became more picky (correctly so) about the SQL used in the flyway migration files.

Specifically, the use of `engine = InnoDB` on the table creation in the 1.0 migration file was not compatible with h2. (Not sure why h2 previously allowed that clause.)

# How

Use the new `@EnableTestContainersDatabase` annotation provided by https://github.com/racker/salus-test/pull/7 to activate mysql testcontainers support in place of h2 during unit tests.

# How to test

Existing unit tests

# Depends on

https://github.com/racker/salus-test/pull/7


src/test/java/com/rackspace/salus/telemetry/entities/JsonClobTypeTest.java
```diff
@@ -77,6 +79,7 @@ public void testNullClobValue() {
         entityManager.persistFlushFind(resource);
 
     assertThat(saved.getId()).isNotNull();
-    assertThat(saved).isEqualToIgnoringGivenFields(resource, "id");
+    assertThat(saved).isEqualToIgnoringGivenFields(resource,
```
harageth: I really like these Hamcrest `assertThat` functions

itzg: Actually these are AssertJ. It’s confusing because there’s Hamcrest and Mockito matchers too.

PR #113, opened 2020-02-10
Move KafkaMessageKey to model from common

jjbuchan:
# What

Moves things between salus-common and this repo.

# How

`KafkaMessageKey` & `KafkaMessageKeyBuilder` move here.
`View` moves over to salus-common

# Why

Initially it was so I could pull `model` into `salus-common` but in the end I have been able to make it so neither repo depends on the other.

`KafkaMessageKey` is only used on model classes so it makes sense to move it here.
`View` isn't really a "model" and will be used by the new role based auth stuff in common so it is moving there to simplify things.


harageth:
This was just a straight copy right?

jjbuchan:
> This was just a straight copy right?

Other than the pom, yeah.  Refactor->Move Class

PR #116, opened 2020-02-21
added a JsonToLowerCaseDeserializer

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-694

# What

Make sure that certain string values are lower case when we deserialize them from json

# How

This provides a class annotation that we can apply on a per field level to tell spring to use this deserializer. It will lowercase any of the fields that it annotates

## How to test

tests are created in resource management

# Why

The current set of PR's is just for resourceId but I can see this being used in the future on other fields


src/main/java/com/rackspace/salus/telemetry/translators/JsonToLowerCaseDeserializer.java
```diff
@@ -0,0 +1,38 @@
...
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+import java.io.IOException;
+
+public class JsonToLowerCaseDeserializer extends StdDeserializer<String> {
+
+
+  //private static final long serialVersionUID = 7527542687158493910L;
```
itzg: Could go ahead and uncomment since StdDeserializer appears to implement Serializable.

harageth: Didn't push up that deletion

harageth:
@itzg how do you feel about the package I placed this in?

itzg:
> @itzg how do you feel about the package I placed this in?

I'm glad you asked since I didn't notice the package before. `translators` doesn't seem like the right place; perhaps `model` instead.

PR #117, opened 2020-03-02
Renamed consecutiveCount to stateDuration

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-824

# What

Makes sure that even if not provided the user still has a consecutive count

# How

It sets the value in the LevelExpression

## How to test

Run the unit tests

# Why

This seemed like the best way to have the value go through the same validations as everything else the user would input.

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/entities/EventEngineTaskParameters.java
+++ b/src/main/java/com/rackspace/salus/telemetry/entities/EventEngineTaskParameters.java
@@ -56,7 +56,7 @@ public class EventEngineTaskParameters {
 
     @Valid
     Expression expression;
-    Integer consecutiveCount = 1;
+    Integer stateDuration = 1;
   }
 
   @Data
--- a/src/main/java/com/rackspace/salus/telemetry/entities/EventEngineTaskParameters.java
+++ b/src/main/java/com/rackspace/salus/telemetry/entities/EventEngineTaskParameters.java
@@ -56,7 +56,7 @@ public class EventEngineTaskParameters {
 
     @Valid
     Expression expression;
-    Integer consecutiveCount;
+    Integer consecutiveCount = 1;
   }
 
   @Data
--- a/src/main/java/com/rackspace/salus/telemetry/entities/EventEngineTaskParameters.java
+++ b/src/main/java/com/rackspace/salus/telemetry/entities/EventEngineTaskParameters.java
@@ -56,7 +56,7 @@ public class EventEngineTaskParameters {
 
     @Valid
     Expression expression;
-    Integer stateDuration = 1;
+    Integer stateDuration;
   }
 
   @Data


src/main/java/com/rackspace/salus/telemetry/entities/EventEngineTaskParameters.java
```diff
@@ -56,7 +56,7 @@
 
     @Valid
     Expression expression;
-    Integer consecutiveCount;
+    Integer consecutiveCount = 1;
```
harageth: This is going to be in minutes. 1 is probably to small of a number. I figured something closer to 3 or 5 would be what we ended up on.

itzg: Hmm, why did we continue to call it consecutive **count**? While you're in here I'd prefer we call this "stateDuration" to borrow from Kapacitor's terminology.

itzg: IMHO, null should be default since I think we should have immediate transitions triggering the event by default.

harageth: This should definitely be renamed to stateDuration. 

I am not sure that this should actually default to immediate transition triggering. I guess flapping detection is setup by default in the tick script builder, so we shouldn't have to worry as much about immediate state changes causing flapping tickets for support.

itzg: I still don't agree, but I can be easily swayed by what others prefer.

GeorgeJahad: I don't feel strongly at all, and I'm always hesitant to disagree with geoff.

itzg: To clarify, I would prefer that the simpler event config be the default and user opt into the more-complicated-to-explain state duration logic; however...I do see merit in protecting the user (or really us) from flapping events in the default case.

harageth: Looking a little closer at our flapping detection it is not turned on by default. There is a boolean the customer can set and if its set they can't change the values so those values are created by default.

jjbuchan: I think it should be null here and any defaults should be handled by policy mgmt.

src/main/java/com/rackspace/salus/telemetry/entities/EventEngineTaskParameters.java
```diff
@@ -56,7 +56,7 @@
 
     @Valid
     Expression expression;
-    Integer stateDuration = 1;
+    Integer stateDuration;
```
itzg: ...meanwhile, over on https://github.com/racker/salus-event-engine-management/pull/39/files#r387412069 I realized I was actually getting myself confused. If you change this field to be an `int`, the default will be zero and that will both be handled by existing task generator logic and the default the user experience to the "immediate" mode of event trigger.

harageth: And the extra funny part about this is that I am pretty sure I made it an `Integer` in the first place to allow for `null` as an applicable value.

PR #118, opened 2020-03-11
Add order to translations

jjbuchan:
# What

Adds an `order` field to monitor tanslations.

# How

Add one new field.

## How to test

Tests in monitor management

# Why

Now for each monitor that requires translations to be applied in a certain order, we can add a value to each translation to determine what order they are performed in.

> Assuming we have a translation that looks for `timeout` and appends an `s` to the end of the value...
>
> If this value is provided via the api `timeout: 2` and the end result in telegraf needs to be `responseTimeout: 2s`, if the key was renamed before the value had the `s` appended the value translation would currently fail because `timeout` could not be found.

> Another more complex scenario - these three translators get called and need to be done in this order:
>
> JoinHostPortTranslator
> ScalarToArrayTranslator
> RenameTypeTranslator

# TODO

Update data-content-loader payloads.

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/entities/MonitorTranslationOperator.java
+++ b/src/main/java/com/rackspace/salus/telemetry/entities/MonitorTranslationOperator.java
@@ -16,7 +16,6 @@
 
 package com.rackspace.salus.telemetry.entities;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
 import com.rackspace.salus.telemetry.model.AgentType;
 import com.rackspace.salus.telemetry.model.ConfigSelectorScope;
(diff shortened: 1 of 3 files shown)


jjbuchan:
I always forget about model tests.  After running them there was a failure due to not adding the `order` column to our sql migrations.

After that things would still fail since `order` is a reserved word.  If we wanted to continue using it I think we'd need to add `hibernate.globally_quoted_identifiers=true` to any service using the db.

To keep things simpler I left the field name as `order` in the java object but changed it to `priority_order` in the db.

itzg:
> To keep things simpler I left the field name as `order` in the java object but changed it to `priority_order` in the db.

Sounds good to me.

PR #120, opened 2020-03-24
added java based flyway migration

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-857

# What

Its meant to to set the baseline values for a previous schema migration

# How

We use the flyway schema management java interfaces to find the data we are looking for and migrate it leveraging java to do the data transformation.

Of note: Since this migration isn't doing any DDL and is only altering data if it fails flyway can just rollback the transaction.

## How to test

Start any SALUS application and validate that the `priority_order` field is set to `0`, `1`, or `2`.

# Why

Because some of the data was being stored in a longtext as json data it made it difficult to query the database directly to delineate which values needed to be set where. Being able to pull the values into java and update values from here just makes more sense.

# TODO

I have tested locally and it appears to be working correctly.


src/main/java/db/migration/V4_2__set_translator_order.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+  }
+
+
+  private class MigrationTranslator {
+    UUID id;
+    int priority_order;
+  }
+
+  private class TestTranslatorMapper implements RowMapper<MigrationTranslator> {
```
GeorgeJahad: TestTranslatorMapper seems like an odd name.  Is there a better one?

GeorgeJahad:
The approach seems straight forward to me.  I like it!

I don't quite understand the migration logic, so I'll let the others approve it.

src/main/java/db/migration/V4_2__set_translator_order.java
harageth: Yeah sorry I had refactor the other inner class to a better name and forgot to change this one.

src/main/java/db/migration/V4_2__set_translator_order.java
```diff
@@ -0,0 +1,77 @@
...
+      ObjectMapper mapper = new ObjectMapper();
+      JsonNode node = mapper.readTree(resultSet.getBytes("translator_spec"));
+      JsonNode type = node.get("type");
+      if(type.asText().equals("renameFieldKey")) {
+        value.priority_order = 2;
+      }else if(type.asText().equals("goDuration")) {
+        value.priority_order = 1;
+      }else {
+        value.priority_order = 0;
+      }
```
jjbuchan: nit: better use of whitespace in all these statements

src/main/java/db/migration/V4_2__set_translator_order.java
```diff
@@ -0,0 +1,77 @@
...
+import org.springframework.jdbc.datasource.SingleConnectionDataSource;
+import org.springframework.stereotype.Component;
+
+@Component
+public class V4_2__set_translator_order extends BaseJavaMigration {
+
+  @Override
+  public void migrate(Context context) throws Exception {
+    JdbcTemplate jdbcTemplate = new JdbcTemplate(new SingleConnectionDataSource(context.getConnection(), true));
+    jdbcTemplate.execute("UPDATE monitor_translation_operators SET priority_order = 0");
```
jjbuchan: Does this set the value to `0` for all rows?

harageth: Yes

src/main/java/db/migration/V4_2__set_translator_order.java
harageth: I dont know why I keep on doing that.. updated

src/main/java/com/rackspace/salus/telemetry/EnableSalusJpa.java
```diff
@@ -29,6 +30,7 @@
 @Target(ElementType.TYPE)
 @Configuration
 @EntityScan("com.rackspace.salus.telemetry.entities")
+@ComponentScan("db.migration")
```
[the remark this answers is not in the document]
itzg: I'm the one that told him to add that :) What I'm still wondering is if we may want to re-situate that package to something like `com.rackspace.salus.telemetry.dbmigrations`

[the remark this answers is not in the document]
itzg: ...yeah, I just double checked the Spring Boot docs wording:

> To use Java-based callbacks, create one or more beans that implement Callback. Any such beans are automatically registered with Flyway

So, it doesn't require any particular package, it just picks up any beans already registered with Spring. The reason for the explicit `@ComponentScan` is because being a library-package, this doesn't full under the default `@SpringBootApplication` component scan scope of each of our applications.

itzg:
As a side note, here's the Spring Boot package where the Flyway autoconfig magic happens:

https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/autoconfigure/flyway/package-frame.html

specifically, it's the conditionals on the main autoconfig class that activate when flyway is added to the classpath

https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration.html

My general technique for unraveling the mystery is to look for a package under `org.springframework.boot.autoconfigure` named for the technology of interest, like `flyway`, `quartz`, `data.jpa`, ...

src/main/java/db/migration/V4_2__set_translator_order.java
```diff
@@ -0,0 +1,77 @@
...
+import org.springframework.jdbc.core.RowMapper;
+import org.springframework.jdbc.datasource.SingleConnectionDataSource;
+import org.springframework.stereotype.Component;
+
+@Component
+public class V4_2__set_translator_order extends BaseJavaMigration {
+
+  @Override
+  public void migrate(Context context) throws Exception {
+    JdbcTemplate jdbcTemplate = new JdbcTemplate(new SingleConnectionDataSource(context.getConnection(), true));
```
itzg: It feels like there's a better Spring Boot way to do this. The docs are a little vague when they say

> By default, Flyway autowires the (@Primary) DataSource in your context and uses that for migrations. 

but I think that means you can autowire a `DataSource` into your component's constructor and then use that to instantiate the `JdbcTemplate`.

src/main/java/db/migration/V4_2__set_translator_order.java
itzg: You can import the Google Java coding standard into IntelliJ to let it do the proper reformatting for you

https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml

src/main/java/com/rackspace/salus/telemetry/EnableSalusJpa.java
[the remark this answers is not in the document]
harageth: I would be good with moving the package. 
I do like that having it in such a different package differentiates it from the rest of the code in the application, but it does make it seem like less of an integration with spring boot.

src/main/java/db/migration/V4_2__set_translator_order.java
harageth: Admittedly the examples they provide are for Spring and not Spring boot, but the Flyway docs did it exactly the way I did it. That being said it does feel like there must be a better Spring Boot way to do this.

src/main/java/com/rackspace/salus/telemetry/EnableSalusJpa.java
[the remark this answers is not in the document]
itzg: > but it does make it seem like less of an integration with spring boot

If it helps you feel better, Spring Boot is not the type of framework to prescribe specific package names to locate things. It's relying on good ol' Spring component scanning by default from the application class' package.

src/main/java/db/migration/V4_2__set_translator_order.java
itzg: Looking at the Spring Boot autoconfig code I noticed there's `context.getConfiguration().getDataSource()` and that's what the docs are probably describing.

src/main/java/com/rackspace/salus/telemetry/EnableSalusJpa.java
```diff
@@ -29,6 +30,7 @@
 @Target(ElementType.TYPE)
 @Configuration
 @EntityScan("com.rackspace.salus.telemetry.entities")
+@ComponentScan("com.rackspace.salus.telemetry.dbmigrations")
```
jjbuchan: I do like this more.

PR #121, opened 2020-03-27
Add policy id to monitor

jjbuchan:
# What

Adds a `policyId` field to the `Monitor` object.

# How

This will be used by https://github.com/racker/salus-telemetry-monitor-management/pull/162 to track which monitors under a specific tenant are tied to a monitor policy.


src/main/resources/db/migration/common/V4_3__add_policy_id.sql
```diff
@@ -1,5 +1,5 @@
 /*
```
itzg: This is probably something to document in the KB @harageth , but I was expecting our migration versioning to tend to increment the first position since flyway says:

> For most cases a simple increasing integer should be all you need

Then the second position value can be used to indicate migrations that are related to each other especially in the case of SQL (schema) + Java (data) migrations.

So I was expecting this file to be `V5_0__ ...`

jjbuchan: I'll change this. I didn't read the flyway docs but that works for me.

I was treating it like a minor version bump since it was an addition to the schema and not a breaking change, but thinking about it... nothing should ever be a breaking change.

itzg: Yeah, it's not consistent with our Maven and Go versioning, but seems like simpler is better for tracking these migrations.

src/main/resources/db/migration/common/V5_0__add_policy_id.sql
```diff
@@ -14,14 +14,4 @@
...
-import java.util.UUID;
-import lombok.Data;
-
-@Data
-@KafkaMessageKey(properties = {"tenantId", "monitorId"})
-public class PolicyMonitorUpdateEvent {
-  String tenantId;
-  UUID monitorId;
-}
+ALTER TABLE monitors ADD policy_id varchar(255) null;
```
harageth: Why does github think you moved the PolictMonitorUpdateEvent code here and then turned it into a SQL file?

src/main/resources/db/migration/common/V4_3__add_policy_id.sql
harageth: I will add a versioning portion to the kb

src/main/resources/db/migration/common/V5_0__add_policy_id.sql
itzg: Yeah, it's strange it (or git) calculated the change like that.

PR #122, opened 2020-03-30
Added entity and repository for envoy tokens

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-846

# What

Needed to add the JPA bits for DB storage of the new Envoy token entities:

https://github.com/Rackspace-Segment-Support/salus-docs-internal/blob/master/proposals/envoy-token-for-cert-retrieval/envoy-token-for-cert-retrieval.md

# How

Added the entity, schema migration, and JPA repository interface.

## How to test

Existing unit tests


harageth:
I just realized I should ask whether we should set an optional expiration timestamp with these?

itzg:
> I just realized I should ask whether we should set an optional expiration timestamp with these?

Good question, but I'd prefer to save that for a future enhancement. v1 tokens don't have an expiration https://developer.rackspace.com/docs/rackspace-monitoring/v1/api-reference/agent-token-operations/#create-an-agent-token

src/main/java/com/rackspace/salus/telemetry/entities/EnvoyToken.java
```diff
@@ -0,0 +1,46 @@
...
+  @Id
+  String token;
+
+  String description;
+
+  @Column(name = "tenant_id", nullable = false)
+  String tenantId;
+
+  @CreationTimestamp
+  @Column(name="created_timestamp", nullable = false)
```
jjbuchan: Since the description can be updated we should have an `updated_timestamp` too

itzg: Good point

src/main/java/com/rackspace/salus/telemetry/entities/EnvoyToken.java
```diff
@@ -41,6 +43,11 @@
   @Column(name="created_timestamp", nullable = false)
   Instant createdTimestamp;
 
+  @UpdateTimestamp
+  @Column(name="updated_timestamp")
+  @NonMetadataField
```
jjbuchan: This this was a copy/paste mistake.

itzg: Sure was

src/main/java/com/rackspace/salus/telemetry/entities/EnvoyToken.java
```diff
@@ -31,6 +34,11 @@
 public class EnvoyToken {
 
   @Id
+  @GeneratedValue
+  @Type(type="uuid-char")
+  UUID id;
+
+  @Column(name="token", nullable = false)
```
jjbuchan, line 41: I forget, does `nullable = false` negate the need for `@NotNull`?

itzg: Really `nullable` isn't important any more since we use flyway for schema management and the `NOT NULL` over there is what matters. But I also forget if Hibernate was properly honoring the `@NotNull` at the schema level.

PR #123, opened 2020-03-31
Support the use of monitoring zone policies

jjbuchan:
Some changes to help support the ability to use monitoring zone metadata policies.

I've set zones and metadata to default to empty collections so we don't have to worry about null exceptions everywhere.

Relates to:
https://github.com/racker/salus-telemetry-monitor-management/pull/164
https://github.com/racker/salus-policy-management/pull/31

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/entities/Monitor.java
+++ b/src/main/java/com/rackspace/salus/telemetry/entities/Monitor.java
@@ -61,7 +61,10 @@ import org.hibernate.validator.constraints.NotBlank;
             + "where :metadataKey member of m.monitorMetadataFields"),
     @NamedQuery(name = "Monitor.getTenantsUsingPolicyMetadataInPlugin",
         query = "select distinct m.tenantId from Monitor m join m.pluginMetadataFields "
-            + "where :metadataKey member of m.pluginMetadataFields")
+            + "where :metadataKey member of m.pluginMetadataFields"),
+    @NamedQuery(name = "Monitor.getTenantsUsingZoneMetadata",
+        query = "select distinct m.tenantId from Monitor m left outer join m.zones as z where "
+            + "m.selectorScope = 'REMOTE' and z.size = 0")
 })
 @Data
 public class Monitor implements Serializable {
--- a/src/main/java/com/rackspace/salus/telemetry/entities/Monitor.java
+++ b/src/main/java/com/rackspace/salus/telemetry/entities/Monitor.java
@@ -116,6 +116,7 @@ public class Monitor implements Serializable {
 
     @ElementCollection(fetch = FetchType.EAGER)
     @CollectionTable(name="monitor_zones", joinColumns = @JoinColumn(name="monitor_id"))
+    @NonMetadataField // technically metadata can be used but not in the same way as others
     List<String> zones;
 
     @Column(name="resource_id")
--- a/src/main/java/com/rackspace/salus/telemetry/entities/Monitor.java
+++ b/src/main/java/com/rackspace/salus/telemetry/entities/Monitor.java
@@ -24,6 +24,7 @@ import com.rackspace.salus.telemetry.model.NonMetadataField;
 import java.io.Serializable;
 import java.time.Duration;
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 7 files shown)


src/main/java/com/rackspace/salus/telemetry/entities/Monitor.java
```diff
@@ -60,7 +61,10 @@
             + "where :metadataKey member of m.monitorMetadataFields"),
     @NamedQuery(name = "Monitor.getTenantsUsingPolicyMetadataInPlugin",
         query = "select distinct m.tenantId from Monitor m join m.pluginMetadataFields "
-            + "where :metadataKey member of m.pluginMetadataFields")
+            + "where :metadataKey member of m.pluginMetadataFields"),
+    @NamedQuery(name = "Monitor.getTenantsUsingZoneMetadata",
```
GeorgeJahad, line 65: would it be better to call this getTenantsWithNoZones()?  (It is the cloning process that actually inserts the zone metadata, right?)

jjbuchan: I expect the only use-case for this will be in Policy Management, here - https://github.com/racker/salus-policy-management/blob/f039e0a9a822a9030466b071788a76419fc918f2/src/main/java/com/rackspace/salus/policy/manage/services/MonitorMetadataPolicyManagement.java#L329-L331

so `getTenantsUsingZoneMetadata` is more descriptive to me, but I don't feel strongly if you want me to change it.

> It is the cloning process that actually inserts the zone metadata

Not quite.  Zone Metadata is only used when binding a monitor that is configured with null/empty zones.  This could be when it is cloned or when it is created for the first time.

Zone Metadata is unique in that it varies based on the resource it is bound to (so cannot be set on the Monitor object), whereas all other metadata is tied directly to the Monitor object (by replacing the previous `null` values on the monitor itself).

E.g.  A monitor using metadata values for `interval` and `zones` will look like this (note the zone is still empty)

```
{
      "id": "42059e48-81ff-43c2-ab65-c3acfa0d50ff",
      "name": null,
      "labelSelector": {},
      "labelSelectorMethod": "AND",
      "resourceId": null,
      "excludedResourceIds": [],
      "interval": "PT1M",
      "details": {
        "type": "remote",
        "monitoringZones": [],
        "plugin": {
          "type": "ping",
          "target": "127.0.0.1",
          "count": 5,
          "pingInterval": null,
          "timeout": null,
          "deadline": null
        }
      },
      "createdTimestamp": "2020-04-13T17:45:05Z",
      "updatedTimestamp": "2020-04-13T17:45:05Z",
      "summary": {
        "target": "127.0.0.1"
      }
    }
```

GeorgeJahad: > getTenantsUsingZoneMetadata is more descriptive to me

It is more descriptive.  It's just that the description is seems misleading to me, and I usually prefer "less" descriptive to misleading.  But if it seems fine to you, then I don't feel strongly enough to request a change, (because it isn't pervasive.)

GeorgeJahad: Just to be clear, the thing that is misleading to me here is the word "Using".  In the getTenantsUsingPolicyMetadataInPlugin query, "Using" means the query is getting the key as a parameter.  Here "Using" means return that tenants which we expect to use zoneMetadata.  No zone metadata is passed in as a parameter.

again, I don't feel strongly enough to request a change.  Just wanted to be clear about what i thought is inconsistent.

jjbuchan: I see what you mean.  I'll try think of the options.

jjbuchan: Since you don't care too much I'm going to leave it as it is.  I'm struggling to think of a better clear wording that isn't a whole sentence.

To me, for all the queries "Using` refers to the fact the tenant is making use of the item in the query name (PolicyMetadataInMonitor / PolicyMetadataInPlugin / ZoneMetadata).  I did not mean "Using" to say that the query requires a parameter to be passed to it.  For that I'd probably follow what the repository methods do and say `By` instead of `Using` - `getTenantsByZoneMetadata`.

`getTenantsWithNoZones` is confusing to me, since tenant's don't have zones, monitors do.  And some monitors under a tenant may have zones set and some may not.

As usual, naming things turns out to be the hardest thing.

PR #124, opened 2020-04-08
Add tenant verification handling

jjbuchan:
# What

Adds the ability to verify a tenant exists in the database before any API operation can be performed on it.

# How

Any service annotated with `@EnableTenantVerification` will have its API requests pass through `TenantVerification.preHandle` before they are passed on to the appropriate handler.

If the `X-Tenant-Id` header is set and the value cannot be found within the `tenant_metadata` table then a `NotFoundException` is thrown and a 404 will be returned to the user.  If the value is found in the database the request will proceed as normal.

Additionally, the db operation to see if a tenant exists will be cached to limit the number of db lookups required.

## How to test

Manually and with new tests being added to each service that utilizes the logic.

# Why

To ensure policies are being bound to tenants correctly we need to ensure we know what type of account the tenant is prior to any resources/monitors etc. being created.

This will also allow us to use the `tenant_metadata` table to determine what tenants are affected by policy changes - currently we get the distinct list of tenants from the `resources` table.

This change will allow policies to be activated for a tenant prior to them creating any resources.  In other words, the customer and their support team will know what monitoring is in effect on the account prior to their devices going online (it would often be the case that monitoring is configured prior to the device going online so it is actively monitored immediately once it changes to an online status).

# Other Thing

My concern about "breaking" all our Insomnia requests due to the requirement of having the `X-Tenant-Id` in the requests for this to be processed turned out to not be a problem.  Since it is safe to ignore tenant verification if there is not a tenant id header set, our existing insomnia configs will continue to work fine - just like any service-to-service requests would work.

Similar to the role-based views, if you want to test this logic you either have to manually set the required headers or use an environment that utilized repose and identity authentication.

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/repositories/TenantMetadataRepository.java
+++ b/src/main/java/com/rackspace/salus/telemetry/repositories/TenantMetadataRepository.java
@@ -25,7 +25,9 @@ import org.springframework.data.repository.PagingAndSortingRepository;
 public interface TenantMetadataRepository extends PagingAndSortingRepository<TenantMetadata, UUID> {
   Optional<TenantMetadata> findByTenantId(String tenantId);
 
-  @Cacheable(cacheNames = "tenant_ids", key = "#tenantId")
+  @Cacheable(cacheNames = "tenant_ids", key = "#tenantId",
+      // avoid caching tenants that don't exist yet
+      unless = "#result == false")
   boolean existsByTenantId(String tenantId);
 }
 
--- a/src/main/java/com/rackspace/salus/telemetry/repositories/TenantMetadataRepository.java
+++ b/src/main/java/com/rackspace/salus/telemetry/repositories/TenantMetadataRepository.java
@@ -19,9 +19,13 @@ package com.rackspace.salus.telemetry.repositories;
 import com.rackspace.salus.telemetry.entities.TenantMetadata;
 import java.util.Optional;
 import java.util.UUID;
+import org.springframework.cache.annotation.Cacheable;
 import org.springframework.data.repository.PagingAndSortingRepository;
 
 public interface TenantMetadataRepository extends PagingAndSortingRepository<TenantMetadata, UUID> {
   Optional<TenantMetadata> findByTenantId(String tenantId);
+
+  @Cacheable(cacheNames = "tenant_ids", key = "#tenantId")
+  boolean existsByTenantId(String tenantId);
 }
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/rackspace/salus/telemetry/repositories/TenantMetadataRepository.java
```diff
@@ -19,9 +19,13 @@
...
 import java.util.Optional;
 import java.util.UUID;
+import org.springframework.cache.annotation.Cacheable;
 import org.springframework.data.repository.PagingAndSortingRepository;
 
 public interface TenantMetadataRepository extends PagingAndSortingRepository<TenantMetadata, UUID> {
   Optional<TenantMetadata> findByTenantId(String tenantId);
+
+  @Cacheable(cacheNames = "tenant_ids", key = "#tenantId")
+  boolean existsByTenantId(String tenantId);
```
itzg, line 29: I seem to remember I tried adding `@Cacheable` to a repository method and the caching system didn't actually intercept it. Did it seem to be working for you? I could have had other things misconfigured at the time.

src/main/java/com/rackspace/salus/telemetry/web/TenantVerification.java
```diff
@@ -0,0 +1,52 @@
...
+
+import com.rackspace.salus.telemetry.model.NotFoundException;
+import com.rackspace.salus.telemetry.repositories.TenantMetadataRepository;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.springframework.web.servlet.handler.HandlerInterceptorAdapter;
+
+public class TenantVerification extends HandlerInterceptorAdapter {
+
+  public static final String ERROR_MSG = "Tenant must be created before any operations can be performed on it";
```
itzg, line 27: Perhaps this wording, just clarify its the objects of the tenant that are being validated
```suggestion
  public static final String ERROR_MSG = "Tenant must be created before any operations can be performed with it";
```

src/main/java/com/rackspace/salus/telemetry/repositories/TenantMetadataRepository.java
jjbuchan: Verified things work by enabling `show-sql`.

`Hibernate: select tenantmeta0_.id as col_0_0_ from tenant_metadata tenantmeta0_ where tenantmeta0_.tenant_id=? limit ?` is ran for anything that is not in the cache.

I also added a commit to stop us from caching tenants that don't exist yet.  Any request for it should check the database to see if it exists up until it does exist, then we can trust the cached response.

PR #130, opened 2020-04-24
added new paged query to find list of ids

harageth:
Connected to https://github.com/racker/salus-telemetry-resource-management/pull/79

This PR makes sure that we are still querying the database for paged content rather than doing paging in the server itself for the list of resourceId's obtained from the label query.

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/repositories/ResourceRepository.java
+++ b/src/main/java/com/rackspace/salus/telemetry/repositories/ResourceRepository.java
@@ -40,6 +40,5 @@ public interface ResourceRepository extends PagingAndSortingRepository<Resource,
 
   List<Resource> findAllByTenantIdAndPresenceMonitoringEnabled(String tenantId, boolean presenceMonitoringEnabled);
 
-  @Query("from Resource r where r.id in :resourceIds")
-  Page<Resource> findByResourceId(List<Long> resourceIds, Pageable page);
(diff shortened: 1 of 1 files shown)


src/main/java/com/rackspace/salus/telemetry/repositories/ResourceRepository.java
```diff
@@ -37,4 +39,7 @@
   Optional<Resource> findByTenantIdAndResourceId(String tenantId, String resourceId);
 
   List<Resource> findAllByTenantIdAndPresenceMonitoringEnabled(String tenantId, boolean presenceMonitoringEnabled);
+
+  @Query("from Resource r where r.id in :resourceIds")
+  Page<Resource> findByResourceId(List<Long> resourceIds, Pageable page);
```
itzg, line 44: I couldn't find [where in the docs](https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#repositories.query-methods.details) this is described, but the phrase "In" in the method name allows for the equivalent of what you handcrafted in the query. IntelliJ is really good about auto-suggesting the JPA query method options:

image
```suggestion
  Page<Resource> findByResourceIdIn(List<Long> resourceIds, Pageable page);
```

harageth: So this is what took me so long to try and find because I was fairly certain it had to exist in the magic function name header but like you said I couldn't find it in the documentation.

PR #131, opened 2020-04-24
made sure the magic function references the correct Resource parameter

harageth:
Merged the last PR in because it worked but didn't realize it was referencing the wrong parameter.


harageth:
> Oh yeah. My mind was in monitor management mode and the "resourceId" reference made sense in that context :)

To be fair resourceId IS a field of Resource. It should just be renamed to resourceName

PR #132, opened 2020-04-30
Ensure metadata is not stored as null

jjbuchan:
Updates resource metadata to be non-null.

I copied the sql migration files from `V3_0__` which shows different queries are needed for h2 vs. mysql.

Also added a java migration file to automatically update the current null values to be an empty list.  I've tested this change locally.

### DIFF
--- a/src/main/resources/db/migration/mysql/V7_0__change_resource_metadata.sql
+++ b/src/main/resources/db/migration/mysql/V7_0__change_resource_metadata.sql
@@ -1,0 +1,19 @@
+/*
+ * Copyright 2020 Rackspace US, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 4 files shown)


src/main/resources/db/migration/mysql/V7_0__change_resource_metadata.sql
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+alter table resources
+    change column metadata
+        metadata longtext not null;;
```
GeorgeJahad, line 19: two ;'s?

jjbuchan: fixed

PR #136, opened 2020-06-01
Get all tenants from tenant db instead of resource db

jjbuchan:
# Fixes

https://jira.rax.io/browse/SALUS-872

# What

Rather than relying on the Resources table to determine what tenants exist, this instead pulls the info from the tenant-metadata table.

This means when a new monitor policy is created it can impact a tenant that exists in the system but does not yet have any resources.

i.e. the policy monitor will be cloned to that tenant but there will not yet be any bound monitors due to no resources existing.


jjbuchan:
@GeorgeJahad there is already a PR for that here https://github.com/racker/salus-policy-management/pull/39

PR #137, opened 2020-06-08
Allow values other than int to be set for a task threshold

jjbuchan:
# What

Goes with https://github.com/racker/salus-event-engine-management/pull/47.

Allows thresholds to be set to values other than Number, e.g. Strings.

No SQL changes are needed since this value gets stored as json

```
  @Type(type = "json")
  @Column(nullable = false)
  EventEngineTaskParameters taskParameters;
```


jjbuchan:
From our conversations today it's a good chance this'll soon become irrelevant but I'm gonna merge anyway.  At least the new test in the related PR will still be useful.

PR #139, opened 2020-06-30
Support logical operations in task criteria

jjbuchan:
# What

Updates task expressions so that they can consist of logical or comparison expressions.

## How to test

See new event management pr - https://github.com/racker/salus-event-engine-management/pull/49

# Why

Allows us to migrate over all ele alarms, such as:

```
if (code != 200) {
  return CRITICAL;
}
if (body regex "error" != "") {
  return CRITICAL;
}
return OK;
```

which can now be combined to effectively be represented like

```
if (code != 200 || body regex "error" != "") return CRITICAL
```

This is needed since TickScript can only have one statement per task state.  This expands on ele's functionality by allowing nested logical expressions like

```
if ((code != 200 || code !=500) && body == "") {
  return CRITICAL;
}
```


src/main/java/com/rackspace/salus/telemetry/entities/EventEngineTaskParameters.java
```diff
@@ -116,4 +107,47 @@ static public boolean valid(String c) {
...
+    public enum Operator {
+      AND, OR
+    }
+  }
+
+  @Data
+  @EqualsAndHashCode(callSuper = false)
+  public static class ComparisonExpression extends Expression {
+    @ExpressionValidator.ComparatorValidation()
+    String comparator;
```
itzg, line 140: It seems like this should be an enum to help with validation. Rather than forcing enum names in the JSON like "GTE", I wonder if the JSON encoding could use something like ">="?

jjbuchan: I agree.  I had started changing it yesterday then thought it wouldn't help so ended up leaving it as is (aka how you did it :) )... but i've taken another look and changed it now.  Getting the right `@Json...` annotation helps.

The one problem is the error message returned to the user is less friendly now.  Since I can't think of an clean solution to that right now I'm gonna ignore it... it can fall under the general error handling jira we have unless something comes to mind.

itzg: Yeah, when I reviewed the other repo I realized the comparator field was a regular string beforehand. Thanks for digging deeper and I agree with cleaning up error messaging during the broader exercise.

jjbuchan:
Pushed one more commit.  I had left the `OK` state in there with the intention of looking into it more...

Now i've looked into it I do see that even though `OK` is a state in kapacitor it is not something you set in tickscript.  If any other states conditions are not met it defaults to OK.

...I'll have to see how this affects us sending custom OK messages but other than that it's all good.

itzg:
> Pushed one more commit. I had left the `OK` state in there with the intention of looking into it more...
> 
> Now i've looked into it I do see that even though `OK` is a state in kapacitor it is not something you set in tickscript. If any other states conditions are not met it defaults to OK.
> 
> ...I'll have to see how this affects us sending custom OK messages but other than that it's all good.

I believe the usage of [critReset](https://docs.influxdata.com/kapacitor/v1.5/nodes/alert_node/#critreset) and others can get pretty close to the asymmetric behavior we always wanted, but default behavior should act like an "OK" transition.

PR #140, opened 2020-07-02
resource and monitor search repository functionality

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-901
https://jira.rax.io/browse/SALUS-902

# What

Make sure that we have functions to search for resources and monitors

# How

Adds functions to the repositories that do the searchs.

## How to test

Unit tests have been added to the respective projects

# Why

I attempted to do the monitor one through inference based on the function name but that quickly became clunky. It was much easier to just write it out as a native query. 

# TODO

Open a new JIRA covering database indexing.

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/repositories/MonitorRepository.java
+++ b/src/main/java/com/rackspace/salus/telemetry/repositories/MonitorRepository.java
@@ -12,6 +12,7 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
+ *
  */
 
 package com.rackspace.salus.telemetry.repositories;
@@ -27,14 +28,11 @@ import org.springframework.data.domain.Page;
 import org.springframework.data.domain.Pageable;
 import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.PagingAndSortingRepository;
+import org.springframework.data.repository.query.Param;
 
 public interface MonitorRepository extends PagingAndSortingRepository<Monitor, UUID> {
 
     boolean existsByIdAndTenantId(UUID id, String tenantId);
-
-    List<Monitor> findByIdIn(List<UUID> ids);
-
-    Page<Monitor> findByIdIn(List<UUID> ids, Pageable pageable);
 
     Optional<Monitor> findByIdAndTenantId(UUID id, String tenantId);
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/rackspace/salus/telemetry/repositories/MonitorRepository.java
```diff
@@ -84,4 +82,9 @@
 
     @Query("select count(m) from Monitor m where m.tenantId = :tenantId and :zone member of m.zones")
     int countAllByTenantIdAndZonesContains(String tenantId, String zone);
+
+    @Query(nativeQuery = true,
+        countQuery     = "SELECT count(*) FROM monitors WHERE tenant_id = :tenantId AND (id LIKE %:searchCriteria% OR monitor_name LIKE %:searchCriteria%) ORDER BY monitor_name",
+        value          = "SELECT * FROM monitors WHERE tenant_id = :tenantId AND (id LIKE %:searchCriteria% OR monitor_name LIKE %:searchCriteria%) ORDER BY monitor_name")
```
jjbuchan, line 88: Cool.  I'm guessing this didn't work with plain jpa queries?

src/main/java/com/rackspace/salus/telemetry/repositories/MonitorRepository.java
```diff
@@ -84,4 +82,9 @@
...
+    Page<Monitor> search(@Param("tenantId")String tenantId, @Param("searchCriteria")String searchCriteria, Pageable pageable);
```
jjbuchan, line 89: Is `@Param` actually needed?  We don't use it in other queries.  Is it just because it's a `nativeQuery`?

jjbuchan: After reading the resource section I wonder if this would work on it's own?

```
findByTenantIdAndIdContainingOrMonitorNameContaining
```

But not sure if it'd be smart enough to realize how the two AND/OR are grouped.

src/main/java/com/rackspace/salus/telemetry/repositories/MonitorRepository.java
itzg: I'm not sure if he tried, but we assumed that JPA wouldn't allow a `LIKE` against the UUID column, `id`; however, MySQL only knows it as a varchar column.

src/main/java/com/rackspace/salus/telemetry/repositories/MonitorRepository.java
itzg: Yeah, I wasn't sure either what grouping it would do with a mix of AND/OR, so had suggested a `@Query`.

src/main/java/com/rackspace/salus/telemetry/repositories/MonitorRepository.java
harageth: No with the id being a UUID as an object we couldn't search for a partial id through a jpa query. Plus it became awkward on the organizational level since we needed the tenantid AND either the id or monitor name to match.

src/main/java/com/rackspace/salus/telemetry/repositories/MonitorRepository.java
harageth: Yes `@Param` is used for native queries for named parameters in the queries. Otherwise the parameters are just numbered. 

And yeah I answered your second question in my previous comment.

PR #143, opened 2020-07-31
added delete tenant by id's to repositories

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-956

# What

Allow us to delete everything for a particular tenant from each repository 

# How

Adds repository functions to delete everything for particular tenants

## How to test

Unit tests added in the specific management services

# Why

This is the simplest way to add database functionality

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/repositories/BoundAgentInstallRepository.java
+++ b/src/main/java/com/rackspace/salus/telemetry/repositories/BoundAgentInstallRepository.java
@@ -44,4 +44,8 @@ public interface BoundAgentInstallRepository extends CrudRepository<BoundAgentIn
   );
 
   List<BoundAgentInstall> findAllByAgentInstall_Id(UUID agentInstallId);
+
+  @Query("from BoundAgentInstall b "
+      + " where b.agentInstall.tenantId = :tenantId" )
+  List<BoundAgentInstall> findAllByTenant(String tenantId);
 }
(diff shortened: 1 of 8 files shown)


src/main/java/com/rackspace/salus/telemetry/repositories/BoundAgentInstallRepository.java
```diff
@@ -44,4 +44,8 @@
   );
 
   List<BoundAgentInstall> findAllByAgentInstall_Id(UUID agentInstallId);
+
+  @Query("from BoundAgentInstall b "
+      + " where b.agentInstall.tenantId = :tenantId" )
+  List<BoundAgentInstall> findAllByTenant(String tenantId);
```
itzg, line 50: You can traverse into the references by method-name-convention; however, perhaps this was a case where the name got ugly. Also, depending on how the agent catalog change looks, you may not need to directly retrieve/delete the bindings.

jjbuchan:
I'm going to wait until the other PRs using this are pushed up.

My first thought is that I didn't expect we'd be doing a `deleteAll` for each entity.  Feels like that would lead to problems if one of them failed?  I'm sure there's workarounds to that but I want to see the code first.

harageth:
> My first thought is that I didn't expect we'd be doing a `deleteAll` for each entity. 

Why wouldn't we want to do a delete all? If the task is to delete everything for a tenant thats what we need to do. I did talk to Geoff about it and thats what he was thinking. The more thoughts about it the better though.

jjbuchan:
All depends on how you're implementing it and how everything ties together.  Since there's various dependencies between things I'd assume with a `deleteAll` approach if there's any failures it would fail early vs. being able to remove as much as possible and only leaving the things that failed to be removed.

I don't know if either way is better than the other, it just wasn't my first thought.

PR #147, opened 2020-08-14
added findMonitorsBoundToResourceAndTenant to BoundMonitorRepository

shabd67:
# Resolves

[SALUS-980](https://jira.rax.io/browse/SALUS-980)

# What

Resolved - Deleting resource also deletes bound monitors for other resources.

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/repositories/BoundMonitorRepository.java
+++ b/src/main/java/com/rackspace/salus/telemetry/repositories/BoundMonitorRepository.java
@@ -108,4 +108,9 @@ public interface BoundMonitorRepository extends CrudRepository<BoundMonitor, Bou
 
   List<BoundMonitor> findAllByMonitor_IdAndZoneNameIn(UUID monitorId, Collection<String> zoneNames);
   List<BoundMonitor> findAllByMonitor_IdAndResourceIdAndZoneNameIn(UUID monitorId, String resourceId, Collection<String> zoneNames);
+
+  @Query("select b from BoundMonitor b"
+      + " where b.resourceId = :resourceId"
+      + " and b.monitor.tenantId = :tenantId")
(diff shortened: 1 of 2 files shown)


src/main/java/com/rackspace/salus/telemetry/repositories/BoundMonitorRepository.java
```diff
@@ -108,4 +108,9 @@
 
   List<BoundMonitor> findAllByMonitor_IdAndZoneNameIn(UUID monitorId, Collection<String> zoneNames);
   List<BoundMonitor> findAllByMonitor_IdAndResourceIdAndZoneNameIn(UUID monitorId, String resourceId, Collection<String> zoneNames);
+
+  @Query("select b from BoundMonitor b"
+      + " where b.resourceId = :resourceId"
+      + " and b.monitor.tenantId = :tenantId")
+  List<BoundMonitor> findMonitorsBoundToResourceAndTenant(String tenantId, String resourceId);
```
itzg, line 115: I like the name of your method. It's a bummer we have the existing one called `findMonitorsBoundToResource` which really should have been called `findMonitorIdsBoundToTenantAndResource`...since it only seems to be used in one non-test location I'd be happy if you renamed that existing one while you're in here.

shabd67: done.

GeorgeJahad: I like to keep names as consistent as possible.  It looks like there is one method name the ends with "TenantAndResource" and one that ends with "ResourceAndTenant".  Seems like they should both end with "TenantAndResource".

shabd67: agreed.

PR #151, opened 2020-09-03
Added pollerResourceId to BoundMonitor and queries to count bindings

itzg:
# Part of

https://jira.rax.io/browse/SALUS-358

# What

Rather than use a brittle, DIY counter in etcd, we'll instead use etcd to simply retrieve a list of the active poller's per remote monitoring zone and use the `BoundMonitor` table as the single source of truth for bindings per envoy-poller.

The motivation to make this change was observing my local system having troubles binding a new monitor to the 1000 resources I had left over from envoy-ambassador stress testing. The `ZoneStorage` code was getting itself into a tight loop of operations like the following with numerous collisions trying to increment the binding count:

```
2020-08-28 13:22:10.821 DEBUG [salus-telemetry-monitor-management,,,] 78771 --- [         etcd22] c.r.s.t.etcd.services.ZoneStorage        : Putting new bound count=894 for resource=development:0 in zone=ResolvedZone(name=dev, tenantId=aaaaaa)
2020-08-28 13:22:10.821 DEBUG [salus-telemetry-monitor-management,,,] 78771 --- [         etcd27] c.r.s.t.etcd.services.ZoneStorage        : Changing bound count of resource=development:0 in zone=ResolvedZone(name=dev, tenantId=aaaaaa) by amount=-1
2020-08-28 13:22:10.827 DEBUG [salus-telemetry-monitor-management,,,] 78771 --- [         etcd30] c.r.s.t.etcd.services.ZoneStorage        : Re-trying incrementing bound count of resource=development:0 in zone=ResolvedZone(name=dev, tenantId=aaaaaa) due to collision
```

# How

Added a nullable `pollerResourceId` column to `BoundMonitor` that is set as a monitor is bound into each specific monitoring zone. (It loosely gets used in conjunction with the pre-existing `envoyId` column, but changes less often since an Envoy's resourceId is stable across re-connects.)

To enable zone allocations (in the PR coming next), a name query has been added each for private zone and public zone where it does a group-by-count query.

## How to test

Added unit test of the new named queries.


GeorgeJahad (approved):

src/main/java/com/rackspace/salus/telemetry/entities/BoundMonitor.java
```diff
@@ -114,6 +140,9 @@
   @Column(name="envoy_id", length = 100)
   String envoyId;
 
+  @Column(name="poller_resource_id", length = 100)
+  String pollerResourceId;
```
jjbuchan, line 144: I had been wondering whether we should store `envoyId` on the Resource and be able to look up any envoy that way.

I've not seen how you're using this yet so maybe that is a different topic.

itzg: Good question, but yes, let's save that for another discussion since these three PRs were massive as it is :)

jjbuchan, line 144: Also slightly wondering whether this value should be returned on BoundMonitorDTO objects for admin authenticated users.

jjbuchan (approved):

src/main/java/com/rackspace/salus/telemetry/entities/BoundMonitor.java
itzg: Ah yes, I forgot to add it over there.

itzg:
Merging now so I can verify build and tests in https://github.com/racker/salus-telemetry-monitor-management/pull/205

PR #153, opened 2020-09-15
Setup Spring JPA Envers audit logging for monitors

harageth:
This PR serves as a basis for how we can setup Spring JPA-Envers style audit logging. This information will be placed in our `docs/proposals` I just wanted a draft PR to reference.

Ultimately this is very simple. You have to tell the base class that it needs to utilize an Envers style factory when instantiating the JPA repositories. 

For each repository we need audited we extend the `RevisionRepository`

We also need to add the `@Audited` annotation with the `@EntityListeners(AuditingEntityListener.class)` annotation to each entity object we want audited. While also adding
```
    @Transient
    private RevisionMetadata<Integer> editVersion;
```
To the class as well. The @CreatedBy annotation shows Spring where to inject the user on create operations. 

Curiously I had set `ddl-auto:true` but accidentally reverted it before I ran this and it still generated the audit tables. 

I do want to point out the fact that the audit table utilizes the same schema as the actual table, and while this is helpful for giving us more granular data that is easier to search through, it also means that when updating the schema of anything in the database the audit table is going to take a long time to update. Part of our data governance strategy should be an understanding of data retention in the audit table and clearing that out. If only to allow us to update the schema more easily. 

# TODO
- [ ] Fix Spring Security to pass through into the microservices so JPA can automatically inject who was doing what.
- [ ] Link to article for injecting not just the authorized user but the impersonation information as well
- [ ] Link to docs showing how to set this up to send data to a different data source
- [ ] Pull the schema information from the audit tables generated by JPA and create flyway migrations out of it

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/entities/Monitor.java
+++ b/src/main/java/com/rackspace/salus/telemetry/entities/Monitor.java
@@ -12,6 +12,7 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
+ *
  */
 
 package com.rackspace.salus.telemetry.entities;
@@ -33,6 +34,7 @@ import javax.persistence.CollectionTable;
 import javax.persistence.Column;
 import javax.persistence.ElementCollection;
 import javax.persistence.Entity;
+import javax.persistence.EntityListeners;
 import javax.persistence.EnumType;
 import javax.persistence.Enumerated;
 import javax.persistence.FetchType;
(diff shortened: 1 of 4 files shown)


itzg:
Some random initial notes:

- That is indeed odd that it still auto created the audit tables since the third/info box at the start of [this section](https://docs.jboss.org/hibernate/orm/current/userguide/html_single/Hibernate_User_Guide.html#envers-basics) says it should only do that when using create/update auto DDL mode. Maybe the previous copy of the file was still lingering in the maven `target` directory.
- I think it would have been nice if they had serialized the entity at each revision to JSON and then stored those in a single, universal audit table by revision. The table-per-entity concept makes sense when the schema is not changing; however, for reasons including the one you mentioned, it doesn't elegantly handle schema migration. For example, an old revision of an entity would be somewhat misleading since it would have to be structured with the newest schema, even though that entity's revision didn't have some of those columns at the time of capture.
- I wonder if we can borrow some of their integration techniques, like the listeners, and do our own schema-migration tolerant version? When retrieving the audit trail of a given entity, I'm not too concerned with retrieving it as a fully blown JPA entity, but rather the JSON-ish summary is sufficient for the audit visualization.
- With that said, I do like the idea of using a first class solution that is part of Hibernate, so the DIY suggestion is not a casual decision to make
- Even without the schema migration envers, it might be time to think about devising a small microservice that's only purpose is to run the flyway migrations. Then we can let that run however long it needs to, applying the migrations, before doing rolling restarts of the applications depending on the migrations.
- I didn't see yet in the Hibernate Envers docs how you can situate those "_AUD" tables in a separate database instance

src/main/java/com/rackspace/salus/telemetry/entities/Monitor.java
```diff
@@ -156,13 +168,27 @@
...
+    @CreatedDate
     Instant createdTimestamp;
 
     @UpdateTimestamp
     @Column(name="updated_timestamp")
     @NonMetadataField
+    @LastModifiedDate
     Instant updatedTimestamp;
 
+    @Column(name = "created_by")
```
jjbuchan, line 180: We should probably consider creating a `BaseEntity` that contains all these standard fields that any other entity can extend.  Another quick google comes up with a `@MappedSuperclass` annotation that may be helpful for this.

It could include id as well as all the creation/update timestamps/users, etc.

PR #154, opened 2020-09-23
added AgentHistory model and its Repo.

shabd67:
# Resolves

[SALUS-796](https://jira.rax.io/browse/SALUS-796)

# What

Added AgentHistory Model and its Repo


src/main/java/com/rackspace/salus/telemetry/repositories/AgentHistoryRepository.java
```diff
@@ -0,0 +1,30 @@
...
+import com.rackspace.salus.telemetry.entities.AgentHistory;
+import java.util.List;
+import java.util.UUID;
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.CrudRepository;
+
+public interface AgentHistoryRepository extends CrudRepository<AgentHistory, UUID> {
+
+  @Query("select a from AgentHistory a where a.envoyID = :envoyId "
+      + "and a.connectedClosedAt IS NULL ORDER BY a.connectedAt DESC")
```
itzg, line 28: You actually didn't need to declare the JPQL here since the [inferred query method lookup](https://docs.spring.io/spring-data/jpa/docs/2.3.4.RELEASE/reference/html/#jpa.sample-app.finders.strategies) will handle these types of conditionals. [This appendix](https://docs.spring.io/spring-data/jpa/docs/2.3.4.RELEASE/reference/html/#repository-query-keywords) further expands upon the phrasing detected by that. I think an "IsNull" would be needed, `findAllByEnvoyIdAAndConnectedClosedAtIsNull`

src/main/java/com/rackspace/salus/telemetry/entities/AgentHistory.java
```diff
@@ -0,0 +1,65 @@
...
+)
+public class AgentHistory {
+
+  @Id
+  @GeneratedValue
+  @org.hibernate.annotations.Type(type="uuid-char")
+  UUID id;
+
+  @Column(name ="connected_timestamp", nullable = false)
+  Instant connectedAt;
```
itzg, line 46: How about simply
```suggestion
  Instant connected;
```

src/main/java/com/rackspace/salus/telemetry/entities/AgentHistory.java
```diff
@@ -0,0 +1,65 @@
...
+
+  @Column(name ="connection_closed_timestamp")
+  Instant connectedClosedAt;
```
itzg, line 49: ...and
```suggestion
  Instant disconnected;
```

src/main/java/com/rackspace/salus/telemetry/entities/AgentHistory.java
```diff
@@ -0,0 +1,65 @@
...
+
+  @Column(name ="tenant_id", nullable = false)
+  String tenantID;
```
itzg, line 52: "ID" is always a weird one...we use strict camel case so these suffixes are always `Id`, such as [these](https://github.com/racker/salus-telemetry-model/blob/aa89a8eb4bb1f807e1becf783a8cf3115b22a87a/src/main/java/com/rackspace/salus/telemetry/entities/BoundMonitor.java#L118-L143)
```suggestion
  String tenantId;
```

src/main/java/com/rackspace/salus/telemetry/entities/AgentHistory.java
```diff
@@ -0,0 +1,65 @@
...
+
+  @Column(name ="resource_id",nullable = false)
+  String resourceID;
```
itzg, line 55:
```suggestion
  String resourceId;
```

src/main/java/com/rackspace/salus/telemetry/entities/AgentHistory.java
```diff
@@ -0,0 +1,65 @@
...
+
+  @Column(name ="envoy_id",nullable = false)
+  String envoyID;
```
itzg, line 58:
```suggestion
  String envoyId;
```

src/main/java/com/rackspace/salus/telemetry/entities/AgentHistory.java
```diff
@@ -0,0 +1,65 @@
...
+
+  @Column(name ="remote_ip",nullable = false)
+  String remoteIP;
```
itzg, line 61:
```suggestion
  String remoteIp;
```

src/main/java/com/rackspace/salus/telemetry/entities/AgentHistory.java
```diff
@@ -0,0 +1,65 @@
...
+
+  @Column(name ="zone_id")
+  String zoneID;
```
itzg, line 64:
```suggestion
  String zoneId;
```

itzg (changes requested):
Any entity changes need flyway migrations added in https://github.com/racker/salus-telemetry-model/tree/master/src/main/resources/db/migration/mysql . I especially wanted to confirm indexes to ensure efficient execution of the repository query.

For research/spike issues we usually write up a proposal in https://github.com/Rackspace-Segment-Support/salus-docs-internal/tree/master/proposals and go through a PR review of that. Your approach looks straight forward, so a proposal probably wasn't needed in this case; however, what are your thoughts about:

- automatic/ongoing pruning of AgentHistory data
- APIs to retrieve agent history. At least a public API is needed

src/main/resources/db/migration/common/V1_0__initial.sql
```diff
@@ -1,3 +1,21 @@
...
+    envoy_id              varchar(255)               not null,
+    remote_ip             varchar(255)               not null,
+    zone_id               varchar(255)               null
+)
+    engine = InnoDB;
+
+create index by_envoy
+    on agent_history (envoy_id);
+
+
```
itzg, line 18: You probably need to read up on Flyway. It's a big no-no to update an existing migration file:

https://flywaydb.org/documentation/migrations#versioned-migrations

Please put new schemas in a new file with an appropriate version identifier.

src/main/resources/db/migration/common/V1_0__initial.sql
```diff
@@ -1,3 +1,21 @@
...
+    tenant_id             varchar(255)               not null,
+    resource_id           varchar(255)               not null,
+    envoy_id              varchar(255)               not null,
+    remote_ip             varchar(255)               not null,
+    zone_id               varchar(255)               null
+)
+    engine = InnoDB;
+
+create index by_envoy
+    on agent_history (envoy_id);
```
itzg, line 16: I think you'll need `tenant_id` added as first column of index. Then it'll support query for all envoy connections by tenant and also by specific envoy.

itzg (changes requested):

src/main/resources/db/migration/common/V1_0__initial.sql
shabd67: will take care of both.

src/main/java/com/rackspace/salus/telemetry/entities/AgentHistory.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+)
+public class AgentHistory {
+
+  @Id
+  @GeneratedValue
+  @org.hibernate.annotations.Type(type="uuid-char")
+  UUID id;
+
+  @Column(name ="connected_timestamp", nullable = false)
```
jjbuchan, line 43: With all of these I wonder if we need a `@NotNull` as well?  @itzg do you remember if there were problems with not including it?

jjbuchan (commented):
Looks good once Geoff's suggestions are included.

While creating this did you notice if there's any other interesting fields we could log from the agent connection?  I'm not sure if there would be anything, just curious.

src/main/resources/db/migration/common/V1_0__initial.sql
shabd67: should I add resourceId also to the columns of index as @jjbuchan recommended to have a public api to get agent connection history with a resourceId?

src/main/java/com/rackspace/salus/telemetry/entities/AgentHistory.java
itzg: It's a bit tedious, but both are actually useful. The `@NotNull` will flag it earlier (and perhaps by IntelliJ). The `nullable = false` will declare the schema itself in MySQL to disallow null values...however, `nullable` is probably ignored since we manage schema via flyway.

src/main/resources/db/migration/common/V1_0__initial.sql
itzg: Broadly, yes; however, it'll require two separate index declarations since the API request would drive queries by tenant+envoyId or tenant+resourceId.

src/main/resources/db/migration/common/V9_2_add_agent_history.sql
```diff
@@ -0,0 +1,16 @@
...
+    tenant_id             varchar(255)               not null,
+    resource_id           varchar(255)               not null,
+    envoy_id              varchar(255)               not null,
+    remote_ip             varchar(255)               not null,
+    zone_id               varchar(255)               null
+)
+    engine = InnoDB;
+
+create index byTenantOrEnvoyId
+    on agent_history (tenant_id, envoy_id);
```
itzg, line 16: Thinking more about [my comment on your proposal doc](https://github.com/Rackspace-Segment-Support/salus-docs-internal/pull/96#discussion_r494360334), I'm wondering if this should be a `create unique index` and your repository method can be simply:

```
Optional<AgentHistory> findByTenantIdAndEnvoyId(String tenantId, String envoyId);
```

src/main/resources/db/migration/common/V9_2_add_agent_history.sql
```diff
@@ -0,0 +1,16 @@
...
+    disconnected_timestamp  datetime                 null,
+    tenant_id             varchar(255)               not null,
+    resource_id           varchar(255)               not null,
+    envoy_id              varchar(255)               not null,
+    remote_ip             varchar(255)               not null,
+    zone_id               varchar(255)               null
+)
+    engine = InnoDB;
+
+create index byTenantOrEnvoyId
```
itzg, line 15: `byTenantOrEnvoyId` is a little misleading of a name. Compound indices require that if the 2nd, 3rd, etc columns are used in a query then all previous columns must be included. So the logical expression would be `by (tenant) or (tenant and envoy)`... but to keep it simple and use lower snakecase, I'd suggest:

```
by_tenant_envoy
```   

then the resourceId index would be named `by_tenant_resource_id`

src/main/java/com/rackspace/salus/telemetry/repositories/AgentHistoryRepository.java
```diff
@@ -0,0 +1,30 @@
...
+
+import com.rackspace.salus.telemetry.entities.AgentHistory;
+import java.util.List;
+import java.util.UUID;
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.repository.PagingAndSortingRepository;
+
+public interface AgentHistoryRepository extends PagingAndSortingRepository<AgentHistory, UUID> {
+  List<AgentHistory> findAllByEnvoyIdAndDisconnectedAtIsNull(String envoyId);
```
itzg, line 27: [This comment](https://github.com/racker/salus-telemetry-model/pull/154#discussion_r495121684) seems to have gotten lost in the index changes.

itzg (commented):
Looks like your flyway file disappeared in latest commit. BTW, it should be named with prefix `V10_0__` since it is not related to the `V9` files.

src/main/java/com/rackspace/salus/telemetry/repositories/AgentHistoryRepository.java
shabd67: `findAllByEnvoyIdAndDisconnectedAtIsNull ` is being used to fetch latest connection for given envoyId and dump it into DB with connectionClosedAt time. 
```
List<AgentHistory> agents = agentHistoryRepository.findAllByEnvoyIdAndDisconnectedAtIsNull(envoyId);
    if(!agents.isEmpty()) {
      AgentHistory agentHistory = agents.get(0);
      final Instant connectionClosedTime = Instant.now();
      agentHistory.setDisconnectedAt(connectionClosedTime);
      agentHistoryRepository.save(agentHistory);
    }
```

src/main/java/com/rackspace/salus/telemetry/repositories/AgentHistoryRepository.java
```diff
@@ -17,14 +17,11 @@
...
-import org.springframework.data.repository.PagingAndSortingRepository;
+import org.springframework.data.repository.CrudRepository;
 
-public interface AgentHistoryRepository extends PagingAndSortingRepository<AgentHistory, UUID> {
-  List<AgentHistory> findAllByEnvoyIdAndDisconnectedAtIsNull(String envoyId);
-  Page<AgentHistory> findByTenantIdAndEnvoyId(String tenantId, String envoyId, Pageable pageable);
-  Page<AgentHistory> findByTenantIdAndResourceId(String tenantId, String resourceId, Pageable pageable);
+public interface AgentHistoryRepository extends CrudRepository<AgentHistory, UUID> {
+  Optional<AgentHistory> findByTenantIdAndEnvoyId(String tenantId, String envoyId);
+  Optional<AgentHistory> findByTenantIdAndResourceIdAndDisconnectedAtIsNull(String tenantId, String resourceId);
```
itzg, line 26: This one still needs to be pageable since it is used via the API and in that case would retrieve any number of envoy connections associated with that resourceId.

Also, why is the "AndDisconnectedAtIsNull" clause now showing up here?

src/main/java/com/rackspace/salus/telemetry/repositories/AgentHistoryRepository.java
```diff
@@ -17,14 +17,11 @@
...
-import org.springframework.data.domain.Page;
-import org.springframework.data.domain.Pageable;
-import org.springframework.data.repository.PagingAndSortingRepository;
+import org.springframework.data.repository.CrudRepository;
 
-public interface AgentHistoryRepository extends PagingAndSortingRepository<AgentHistory, UUID> {
-  List<AgentHistory> findAllByEnvoyIdAndDisconnectedAtIsNull(String envoyId);
-  Page<AgentHistory> findByTenantIdAndEnvoyId(String tenantId, String envoyId, Pageable pageable);
-  Page<AgentHistory> findByTenantIdAndResourceId(String tenantId, String resourceId, Pageable pageable);
+public interface AgentHistoryRepository extends CrudRepository<AgentHistory, UUID> {
```
itzg, line 24: It seems like you also need a query method to address https://github.com/Rackspace-Segment-Support/salus-docs-internal/pull/96#discussion_r494361434

src/main/java/com/rackspace/salus/telemetry/repositories/AgentHistoryRepository.java
shabd67: @itzg , I got your point for "findByTenantIdAndResourceId", but for findByTenantIdAndEnvoyId, will it be Pageable as well? actually I am using same method to update recently closed envoy connection with disconnectedAt column.

src/main/java/com/rackspace/salus/telemetry/repositories/AgentHistoryRepository.java
shabd67: yes, I will be needing one for that.

src/main/java/com/rackspace/salus/telemetry/repositories/AgentHistoryRepository.java
shabd67: while at api side, findByTenantIdAndResourceId and findByTenantId will return a PagedContent while findByTenantIdAndEnvoyId will return an Optional.

itzg: I had started to comment over there, but for all three of the API query types you'll return `PagedContent` (populated from a `Page`). In the case of query by envoyId, use the `PageContent.ofSingleton` to unify that response regardless of single value or paged values.

src/main/java/com/rackspace/salus/telemetry/entities/AgentHistory.java
jjbuchan: @shabd67 annotating the non-null fields with `@NotNull` would be good to do.

jjbuchan (changes requested):

itzg (approved):
Looks good. Please wait on @jjbuchan 's re-review before merging.

jjbuchan (dismissed):

itzg (approved):

jjbuchan (approved):

PR #158, opened 2020-10-13
Remove h2 migrations

jjbuchan:
We no longer need `h2` migrations.  I've consolidated all the required migrations under `/mysql`.

## Why

To remove legacy things and make it less confusing.


itzg (approved):
Thanks. I had been wondering about doing that when I last added a migration and then forgot.

PR #161, opened 2020-10-15
Created new table to track jobs

rohitsngh27:
# Resolves
https://jira.rax.io/browse/SALUS-1051

# What
A new table is created to keep track of the jobs results.

# How
Created new table and repository.

### DIFF
--- a/src/main/java/com/rackspace/salus/telemetry/entities/Jobs.java
+++ b/src/main/java/com/rackspace/salus/telemetry/entities/Jobs.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2020 Rackspace US, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.rackspace.salus.telemetry.entities;
+
+import com.rackspace.salus.telemetry.model.JobName;
+import com.rackspace.salus.telemetry.model.JobStatus;
+import java.io.Serializable;
+import java.time.Instant;
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.EnumType;
+import javax.persistence.Enumerated;
+import javax.persistence.Id;
+import javax.persistence.Table;
+import javax.validation.constraints.NotBlank;
+import lombok.Data;
+import org.hibernate.annotations.CreationTimestamp;
+import org.hibernate.annotations.Type;
+import org.hibernate.annotations.UpdateTimestamp;
+
+@Entity
+@Table
+@Data
+public class Jobs implements Serializable {
+
+  @Id
+  String id;
+
+  @Column(name = "tenant_id")
+  String tenantId;
+
+  @Column
+  @NotBlank
+  @Enumerated(EnumType.STRING)
+  JobName name;
+
+  @Column
+  @NotBlank
+  @Enumerated(EnumType.STRING)
+  JobStatus status;
+
+  @Column
+  @Type(type = "text")
+  String description;
+
+  @CreationTimestamp
+  @Column(name = "created_timestamp")
+  Instant createdTimestamp;
+
+  @UpdateTimestamp
+  @Column(name = "updated_timestamp")
+  Instant updatedTimestamp;
+}
--- a/src/main/java/com/rackspace/salus/telemetry/entities/Jobs.java
+++ b/src/main/java/com/rackspace/salus/telemetry/entities/Jobs.java
@@ -16,7 +16,6 @@
 
 package com.rackspace.salus.telemetry.entities;
 
-import com.rackspace.salus.telemetry.model.JobName;
 import com.rackspace.salus.telemetry.model.JobStatus;
 import java.io.Serializable;
 import java.time.Instant;
@@ -27,13 +26,14 @@ import javax.persistence.Enumerated;
 import javax.persistence.Id;
 import javax.persistence.Table;
 import javax.validation.constraints.NotBlank;
+import javax.validation.constraints.NotNull;
 import lombok.Data;
 import org.hibernate.annotations.CreationTimestamp;
 import org.hibernate.annotations.Type;
 import org.hibernate.annotations.UpdateTimestamp;
 
 @Entity
-@Table
+@Table(name = "jobs")
 @Data
 public class Jobs implements Serializable {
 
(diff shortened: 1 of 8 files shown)


src/main/java/com/rackspace/salus/telemetry/model/JobStatus.java
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.rackspace.salus.telemetry.model;
+
+public enum JobStatus {
+  SUCCESS, FAILURE
```
jjbuchan, line 20: I'd expect an `IN_PROGRESS` as well.  Especially since this field is `@NotNull`.  Something should be inserted when the job starts.

src/main/java/com/rackspace/salus/telemetry/entities/Jobs.java
```diff
@@ -0,0 +1,66 @@
...
+
+  @Id
+  String id;
+
+  @Column(name = "tenant_id")
+  String tenantId;
+
+  @Column
+  @NotBlank
+  String name;
```
jjbuchan, line 48: I think this should either use the enum `JobName` or the `JobName` class should be removed and we just allow any string.

Slightly indifferent to what option is chosen.  I can see benefits to both.

jjbuchan (changes requested):

src/main/java/com/rackspace/salus/telemetry/entities/Jobs.java
rohitsngh27: I am using value of JobName enums to store in database that's why I made it as String here.

src/main/java/com/rackspace/salus/telemetry/entities/Jobs.java
```diff
@@ -0,0 +1,66 @@
...
+import javax.validation.constraints.NotNull;
+import lombok.Data;
+import org.hibernate.annotations.CreationTimestamp;
+import org.hibernate.annotations.Type;
+import org.hibernate.annotations.UpdateTimestamp;
+
+@Entity
+@Table(name = "jobs")
+@Data
+public class Jobs implements Serializable {
```
itzg, line 38: nit: we use singular class names, but plural table names
```suggestion
public class Job implements Serializable {
```

src/main/java/com/rackspace/salus/telemetry/entities/Jobs.java
itzg: > I am using value of JobName enums to store in database that's why I made it as String here.

If sticking with the enum, I prefer database fields to use `@Enumerated(EnumType.STRING)` like `status` rather than a label-type string.

src/main/java/com/rackspace/salus/telemetry/entities/Jobs.java
```diff
@@ -0,0 +1,66 @@
...
+  String name;
+
+  @Column
+  @NotNull
+  @Enumerated(EnumType.STRING)
+  JobStatus status;
+
+  @Column
+  @Type(type = "text")
+  String description;
```
itzg, line 57: Can you add javadoc to this one to remind us what is typical/expected to go in this field?

src/main/java/com/rackspace/salus/telemetry/model/JobName.java
```diff
@@ -0,0 +1,31 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.rackspace.salus.telemetry.model;
+
+public enum JobName {
```
itzg, line 19: Seeing the enum, it seems like this should be called `JobType` and the field in `Job` be renamed to `type`. As `name` it seems like it should be more open-ended and arbitrary, which might be fine too.

src/main/java/com/rackspace/salus/telemetry/repositories/JobsRepository.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.rackspace.salus.telemetry.repositories;
+
+import com.rackspace.salus.telemetry.entities.Jobs;
+import org.springframework.data.repository.CrudRepository;
+
+public interface JobsRepository extends CrudRepository<Jobs, String> {
```
itzg, line 21:
```suggestion
public interface JobRepository extends CrudRepository<Jobs, String> {
```

itzg (changes requested):

src/main/java/com/rackspace/salus/telemetry/entities/Jobs.java
rohitsngh27: So I should remove the values from this enum. Right?

src/main/java/com/rackspace/salus/telemetry/model/JobName.java
itzg: You can keep them if you want for possible API response enrichment or logging

src/main/java/com/rackspace/salus/telemetry/model/JobType.java
```diff
@@ -0,0 +1,31 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.rackspace.salus.telemetry.model;
+
+public enum JobType {
+  DELETE_TENANT("Delete tenant");
```
jjbuchan, line 20:
```suggestion
  DELETE_TENANT("Delete Tenant");
```

I hesitated on being this picky last night, but the lowercase `t` is still bugging me a little :)

jjbuchan (dismissed):
An optional nit

itzg (approved):

src/main/java/com/rackspace/salus/telemetry/model/JobType.java
```diff
@@ -0,0 +1,31 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.rackspace.salus.telemetry.model;
+
+public enum JobType {
+  DELETE_TENANT("Delete Tenant");
```
jjbuchan, line 20: Per my comment in https://github.com/racker/salus-telemetry-monitor-management/pull/221/files

We might want to rename or add a job for `DELETE_TENANT_MONITORS`

rohitsngh27: Yes renaming it will make more sense

itzg (approved):

jjbuchan (approved):

PR #162, opened 2020-10-20
Remove kapacitorId, add partitionId, and create subtypes for event tasks

jjbuchan:
Update: this now includes the code from two other PRs that were branched off of this: https://github.com/racker/salus-telemetry-model/pull/163 & https://github.com/racker/salus-telemetry-model/pull/164

I've updated the PR title to reflect that, but not the description below.

---

# Related PRs

https://github.com/racker/salus-event-engine-management/pull/73

# What

1. Removed kapacitorId from event tasks
1. Removes the irrelevant index (these aren't used since we use migration files for that)
1. Renames the table to make it plural
1. Auto generate an id on save

# Why

Part of the process of switching from Kapacitor to Esper for the event engine.

# TODO

I'm going to create a chain of additional PRs to this one to fully build out the new model while trying to keep each individual one easy(ish) to reivew


itzg (dismissed):

jjbuchan:
@itzg do you happen to remember what your comment on this was?  I don't see anything.

itzg:
> @itzg do you happen to remember what your comment on this was? I don't see anything.

Hmm, I don't remember either if/what the comment was.

src/main/resources/db/migration/mysql/V11_1__add_state_index.sql
```diff
@@ -14,4 +14,4 @@
  * limitations under the License.
  */
 
- CREATE INDEX by_tenant_resource_monitor_task ON state_changes (tenant_id, resource_id, monitor_id, task_id);
\ No newline at end of file
+CREATE INDEX by_tenant_resource_monitor_task ON state_changes (tenant_id, resource_id, monitor_id, task_id);
```
itzg, line 17: I wonder if this whitespace change affects the flyway checksum?

itzg (approved):
...unless the checksum on modified file is an issue

src/main/resources/db/migration/mysql/V11_1__add_state_index.sql
jjbuchan: Locally it didn't.  I'll be sure it double check that again before merging.

jjbuchan:
@itzg your comment does remind me of something.  Prior to deploying all these new changes I propose we get ops to do a delete all on the tasks table in all environments.  The change in tables caused by adding the abstract class makes adding a java migration of existing objects difficult.  Since any existing tasks were probably created via basic testing it shouldn't matter if we wipe them all.

itzg:
> @itzg your comment does remind me of something. Prior to deploying all these new changes I propose we get ops to do a delete all on the tasks table in all environments. The change in tables caused by adding the abstract class makes adding a java migration of existing objects difficult. Since any existing tasks were probably created via basic testing it shouldn't matter if we wipe them all.

Yeah, that seems reasonable.

GeorgeJahad (dismissed):

jjbuchan:
Note: This was previously approved.  Merging the other PRs branched off of it (which were also approved) has cleared it.

Once the other event-mgmt PRs are approved i'll be testing this one last time locally to make sure all migrations act as expected.

itzg (approved):

src/main/resources/db/migration/mysql/V11_1__add_state_index.sql
jjbuchan: I'm too lazy to test again so just put it back.  I did find an old issue that states whitespace is a problem for flyway.

I'm not sure it would affect us either way since we don't have any state changes stored yet... but I'm going with laziness for now.

PR #166, opened 2020-10-27
Add missing foreign keys

jjbuchan:
# What

Adds foreign keys for entities that should use them.

# How

Using [this method](https://github.com/Rackspace-Segment-Support/salus-docs-internal/issues/101) and plucking out the ones that weren't already created.

## How to test

Once all the other existing model PRs are merged I'll run this locally and make sure things start up fine.

# Why

I happened to notice they were missing.  It should help the integrity of our db.


itzg (approved):
Thanks! And nice technique in the KB.

GeorgeJahad (approved):
I'm surprised these aren't there already.  Thanks!

jjbuchan:
All seems fine locally.  After populating a bunch of data into the tables in switched to this branch and restarted services.

PR #167, opened 2020-10-28
Added query method to get all zones for a tenant

rohitsngh27:
# Resolves
https://jira.rax.io/browse/SALUS-992

# What
Added query method to get all zones for a tenant

# How
Added a new query to get all the zones of a given tenant id.


src/main/java/com/rackspace/salus/telemetry/repositories/ZoneRepository.java
```diff
@@ -37,6 +37,9 @@
             + ", name ASC")
     Page<Zone> findAllAvailableForTenant(String tenantId, Pageable page);
 
+    @Query("select z.name from Zone z where z.tenantId = :tenantId")
+    Page<String> findAllZoneNameForTenant(String tenantId, Pageable page);
```
itzg, line 41: This is in the gray area of a named query, which would be declared on the entity, but since it is technically a projection of the zone entities then I think it is good here.

itzg (approved):
Wait for @jjbuchan since I'd like him to double check my logic on repository query vs named query.

src/main/java/com/rackspace/salus/telemetry/repositories/ZoneRepository.java
jjbuchan: I'm fine with this - pretty sure this is how I did things before - but your reasoning in the past was anything that isn't returning the main object (Zone in this case) should be a named query?

itzg: I'm forgetting too what the scenario was before, but it feels like it was something slightly different than a one-to-one projection.

jjbuchan: If you're good with it, I'm good with it.  Approved!

jjbuchan (approved):

PR #171, opened 2020-11-04
Added targetClassName and monitorType to MetadataPolicyEvent

shabd67:
# Resolves

[SALUS-1033](https://jira.rax.io/browse/SALUS-1033)

# What

BugFix - Metadata policy is being cashed even after deletion


itzg (approved):
Might be good to wait on merging these until @jjbuchan is back on Monday.

jjbuchan (approved):

PR #177, opened 2020-12-14
Updated event task data structures for common EventProcessor approach

itzg:
# Part of

https://jira.rax.io/browse/SALUS-1078

# What

The Salus Event Engine is getting completely re-implemented with its own state machine logic, so this PR includes changes that adjust the event task parameters to accommodate that change. It might make more sense to review [that PR](https://github.com/racker/salus-event-engine/pull/1) first or at the same time.


src/main/java/com/rackspace/salus/telemetry/entities/EventEngineTaskParameters.java
```diff
@@ -87,6 +124,28 @@
...
     }
 
+    /**
+     * When importing expressions from external sources the comparison might
+     * have the literal, comparison value on the left. This method will flip
+     * the meaning of the operator to normalize to the comparison value on the
+     * right.
+     * @return the commutative version of this operator
+     */
+    public Comparator flip() {
```
harageth, line 134: nit: This could be moved to be a function on the Comparator. Removing the need for a swtich statement. Personally I like handling control logic on Enums as opposed to if or switch statements.

itzg: Good idea

itzg: and now I remember that I had attempted that, but bumped into

image

and

image

harageth: Ahhh normally how I handle this is just registering a function on each of the enums themselves as opposed to handling something through the constructor. 

I believe something like this would work.
Screen Shot 2021-01-04 at 11 00 51 AM

harageth: Note: PolicyScope was used because it was convenient and was easier than pulling your branch down locally.

itzg: A sub-class for each enum element is both elegant and hacky. BTW, `flip` is already a method of the Comparator enum, so I feel like we're spending way too much time on this one.

harageth: That's why I made the original comment that it was a nit. 

IMO this sort of decision is better for handling state alteration control. It clearly delineates where the state changes are happening (as opposed to never being sure from checking the code) and makes it very clear understanding what states can move to which other states.

itzg: > That's why I made the original comment that it was a nit.

Ah yes, you did.

> IMO this sort of decision is better for handling state alteration control. It clearly delineates where the state changes are happening (as opposed to never being sure from checking the code) and makes it very clear understanding what states can move to which other states.

This flip method doesn't have anything to do with the state processing. FWIW, it's used in this code you wouldn't have reviewed yet:

https://github.com/Rackspace-Segment-Support/salus-migrations/blob/60e3fd760c117d49d2950d60658dddb7edaa8563/ele-salus-adapter/src/main/java/com/rackspace/monplat/elesalusadapter/alarms/TaskParametersVisitor.java#L125

Where that code grammatically normalizes a comparison expression like `50 < idle` to always have the literal on the right, as `idle > 50`.

harageth (dismissed):

rohitsngh27 (approved):

shabd67 (approved):

harageth (approved):

harageth (approved):

rohitsngh27 (approved):

